PR #8, opened 2018-04-04
Debugger local symbol provider

vasily-kirichenko:
image


vasily-kirichenko:
I've rebased on master and removed the unneeded code as you updated FCS.

PR #11, opened 2018-06-12
Use rdgen Gradle plugin to generate models

yole:


rider-fsharp/build.gradle.kts
```diff
@@ -32,7 +49,7 @@ intellij {
     // This will download from www.jetbrains.com/intellij-repository/snapshots or
     // www.jetbrains.com/intellij-repository/releases, respectively.
     // Note that there's no guarantee that these are kept up to date
-    // version = 'LATEST-TRUNK-SNAPSHOT'
+     version = "LATEST-EAP-SNAPSHOT"
```
shalupov: There is no LATEST-EAP-SNAPSHOT version of Rider. Only LATEST-TRUNK-SNAPSHOT or XXXX.Y-SNAPSHOT, see https://www.jetbrains.com/intellij-repository/snapshots/

yole: I didn't mean to change this. Reverted.

PR #53, opened 2019-07-04
Support members in type code vision

vasily-kirichenko:


ReSharper.FSharp/test/data/features/daemon/inferredTypeCodeVision/Common.fs.gold
```diff
@@ -0,0 +1,6 @@
+﻿module Module
+
+let |x|(0) = 1
+
+---------------------------------------------------------
+(0): CodeInsights:
```
vasily-kirichenko: @auduchinok Here I expect something like "(0): CodeInsights: **int**", but it's empty. I have no idea how to fix this.

auduchinok: I'll look into it, thanks!

auduchinok:
@vasily-kirichenko Thanks, I rebased and merged it to `192`.

ReSharper.FSharp/src/FSharp.Psi.Features/src/Daemon/Stages/InferredTypeCodeVisionProvider.fs
```diff
@@ -0,0 +1,102 @@
...
+              match groups with
+              | [[]] when mfv.IsMember -> [["unit"]]
+              | _ -> groups
+
+        [ for group in parameterGroups do
+            let groupStr = group |> String.concat " * "
+            if group.Length = 1 then yield groupStr
+            else yield sprintf "(%s)" groupStr
+          yield returnTypeStr ]
+        |> String.concat " -> "
```
drhumlen: Is it possible to add formatting here? Like colorizing the type names different from symbols ("like ->").

Also, font ligatures doesn't seem to work inside code vision. Perhaps the unicode arrow would be better here?

Last question, perhaps for @auduchinok : can you change the font/size/weight of code vision? It hurts my OCD that the font isn't monospaced xD

vasily-kirichenko: This is how it looks like with a unicode arrow:

image

PR #98, opened 2020-03-04
Add indenting formatter rules for bindings and nested module declarations

reacheight:
Added some formatter rules for indents inside
- bindings
- nested modules

Also added registry value to enable formatter.

### DIFF
--- a/ReSharper.FSharp/src/FSharp.Common/src/Util/RegistryUtil.fs
+++ b/ReSharper.FSharp/src/FSharp.Common/src/Util/RegistryUtil.fs
@@ -33,19 +33,14 @@ type ProtocolSolutionExtensions =
         with _ -> null
 
 [<AbstractClass; Sealed; Extension>]
-type FSharpExperimentalFeaturesEx() =
-    static let getFsModelFlagIfNotEnabled property enabled (solution : ISolution) =
-        if enabled then true else
+type FSharpExperimentalFeaturesEx =
+    [<Extension>]
+    static member FSharpExperimentalFeaturesEnabled(solution: ISolution) =
+        if FSharpRegistryUtil.AllowExperimentalFeaturesCookie.Enabled then true else
 
         match solution.RdFSharpModel() with
         | null -> false
-        | fsModel -> property fsModel
-        
-    [<Extension>]
-    static member FSharpExperimentalFeaturesEnabled(solution: ISolution) =
-        getFsModelFlagIfNotEnabled (fun fsModel -> fsModel.EnableExperimentalFeatures.Value)
-            FSharpRegistryUtil.AllowExperimentalFeaturesCookie.Enabled
-            solution
+        | fsModel -> fsModel.EnableExperimentalFeatures.Value
 
     [<Extension>]
     static member FSharpExperimentalFeaturesEnabled(node: ITreeNode) =
@@ -53,9 +48,11 @@ type FSharpExperimentalFeaturesEx() =
 
     [<Extension>]
     static member FSharpFormatterEnabled(solution: ISolution) =
-        getFsModelFlagIfNotEnabled (fun fsModel -> fsModel.EnableFormatter.Value)
-            FSharpRegistryUtil.AllowFormatterCookie.Enabled
-            solution
+        if FSharpRegistryUtil.AllowFormatterCookie.Enabled then true else
+
+        match solution.RdFSharpModel() with
+        | null -> false
+        | fsModel -> fsModel.EnableFormatter.Value
 
     [<Extension>]
     static member FSharpFormatterEnabled(node: ITreeNode) =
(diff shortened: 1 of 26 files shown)


rider-fsharp/src/main/resources/META-INF/plugin.xml
```diff
@@ -84,6 +84,7 @@
     </intentionAction>
 
     <registryKey key="rider.fsharp.experimental" description="Enable experimental F# features" defaultValue="false" restartRequired="false"/>
+    <registryKey key="rider.fsharp.formatter" description="Enable new F# code formatter" defaultValue="false" restartRequired="false"/>
```
auduchinok: We didn't have a working R# formatter previously, and Fantomas is doing a somewhat different thing in Reformat Code action, so I'd prefer it without `new` word here. Though, I doubt many people will ever see this temporary setting. :)

ReSharper.FSharp/src/FSharp.Psi/src/CodeFormatter/FSharpCodeFormatterInfoProvider.cs
```diff
@@ -12,8 +15,78 @@ public class FSharpFormatterInfoProvider :
...
+        ("ForEachExpr", ElementType.FOR_EACH_EXPR, ForEachExpr.DO_EXPR),
+        ("WhileExpr", ElementType.WHILE_EXPR, WhileExpr.DO_EXPR),
+        ("DoExpr", ElementType.DO_EXPR, DoExpr.EXPR),
+        ("AssertExpr", ElementType.ASSERT_EXPR, AssertExpr.EXPR),
+        ("LazyExpr", ElementType.LAZY_EXPR, LazyExpr.EXPR),
+        ("FixedExpr", ElementType.FIXED_EXPR, FixedExpr.EXPR),
+        ("InferredUpcastExpr", ElementType.INFERRED_UPCAST_EXPR, InferredUpcastExpr.EXPR),
+        ("InferredDowncastExpr", ElementType.INFERRED_DOWNCAST_EXPR, InferredDowncastExpr.EXPR),
+        ("ComputationExpr", ElementType.COMPUTATION_EXPR, ComputationExpr.EXPR),
+        ("LambdaExpr", ElementType.LAMBDA_EXPR, LambdaExpr.EXPR),
```
auduchinok: Does it support deindenting like in the following code?

```fsharp
let foo x = x |> fun _ ->
    ()
```

ReSharper.FSharp/src/FSharp.Psi/src/CodeFormatter/FSharpCodeFormatterInfoProvider.cs
```diff
@@ -12,8 +15,78 @@ public class FSharpFormatterInfoProvider :
...
+      var bindingAndModuleDeclIndentingRulesParameters = new[]
+      {
+        ("NestedModuleDeclaration", ElementType.NESTED_MODULE_DECLARATION, NestedModuleDeclaration.MODULE_MEMBER),
+        ("TopBinding", ElementType.TOP_BINDING, TopBinding.CHAMELEON_EXPR),
+        ("LocalBinding", ElementType.LOCAL_BINDING, LocalBinding.EXPR),
+      };
+
+      var synExprIndentingRulesParameters = new[]
+      {
+        ("ForExpr", ElementType.FOR_EXPR, ForExpr.DO_EXPR),
```
auduchinok: Would be great to have some simple tests like checking if the following code is unchanged.

```fsharp
for _ in () do
    ()
```

ReSharper.FSharp/src/FSharp.Psi/src/CodeFormatter/FSharpCodeFormatterInfoProvider.cs
reacheight: Yes.

ReSharper.FSharp/src/FSharp.Common/src/Util/RegistryUtil.fs
```diff
@@ -8,19 +8,22 @@ open JetBrains.ReSharper.Psi.Tree
 open JetBrains.Rider.Model
 
 module FSharpRegistryUtil =
-    type AllowExperimentalFeaturesCookie() =
+    [<AbstractClass>]
+    type EnabledCookieBase<'T when 'T :> EnabledCookieBase<'T> and 'T : (new : unit -> 'T)>() =
         static let mutable enabled = false
```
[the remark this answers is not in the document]
reacheight: It's unique for every combination of type arguments, when the base class is generic.
изображение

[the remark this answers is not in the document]
auduchinok: OK, I see, thanks!

PR #127, opened 2020-05-24
Show background task in UI when FCS reactor is busy

saul:
loading-file-2

**Todo**:

- [x] Only show in internal mode/have a longer delay before showing
- [ ] Integrate properly with FCS instead of regex matching on trace messages


ForNeVeR:
> **Todo**:
> 
> * [ ]  Hide in internal mode/[…]

Probably you meant to hide it _outside_ of internal mode? I do believe this is useful in regular mode (probably with longer delay, as you suggest).

saul:
I did mean outside of internal mode, yes :) Updated to have a 5 second wait in normal mode and a 1 second delay in internal mode.

PR #134, opened 2020-06-07
Regex language injection

saul:
image

**Todo**:

- [x] Write tests
- [ ] Fix regular string support - escape characters are seemingly ignored
- [x] Support RegExpPattern attribute
- [x] Support completion 
- [x] Support Regex Provider
- [x] Support injection by comment
- [x] Support StringSyntax attribute
- [x] Support RegularExpression attribute
- [ ] Support injection into assignments


ReSharper.FSharp/src/FSharp.Psi.Features/src/Injected/FSharpLiteralInjectionTarget.fs
```diff
@@ -0,0 +1,101 @@
...
+                    ClrRegexLexerOptions(true)
+                | _ ->
+                    ClrRegexLexerOptions.None
+
+            CSharpRegExpBuffer(StringBuffer(text), literalType, lexerOptions) :> _
+
+        override _.DoNotProcessNodeInterior _ = false
+
+        override _.IsPrimaryLanguageApplicable(sourceFile) =
+            sourceFile.LanguageType :? FSharpProjectFileType
```
auduchinok, line 86:
```suggestion
            sourceFile.LanguageType.Is<FSharpProjectFileType>()
```

ReSharper.FSharp/src/FSharp.Psi.Features/src/Injected/FSharpLiteralInjectionTarget.fs
```diff
@@ -0,0 +1,101 @@
...
+
+        override _.AllowsLineBreaks _ = true
+
+        override _.IsWhitespaceToken(token) =
+            token.GetTokenType().IsWhitespace
+
+        // Used only in JS/JSON for now
+        override x.FixValueRangeForLiteral(element) =
+            let startOffset = (x :> IInjectionTargetLanguage).GetStartOffsetForString(element)
+            element.GetTreeTextRange().TrimLeft(startOffset).TrimRight(1)
```
auduchinok, line 99: Do we need this? We're not going to inject js/json on backend AFAIK, is it needed for something else?

ReSharper.FSharp/src/FSharp.Psi/src/Annotations/FSharpAnnotatedCacheProcessor.cs
```diff
@@ -0,0 +1,138 @@
...
+      private readonly AnnotatedEntitiesSet myContext;
+
+      public Processor(HashSet<string> attributeNames, AnnotatedEntitiesSet context)
+      {
+        myAttributeNames = attributeNames;
+        myContext = context;
+      }
+
+      public bool ProcessingIsFinished => false;
+      public bool InteriorShouldBeProcessed(ITreeNode element) => element is not IChameleonNode;
```
auduchinok, line 35: We should probably also cover cases where parameters are inside a chameleon.
```fsharp
let f x =
    fun ([<A>] y) ->
        ()
```
and
```fsharp
let f =
    fun ([<A>] y) ->
        ()
```

That would need an addition to `ChameleonExpressionUtil` too.

ReSharper.FSharp/src/FSharp.Psi/src/Annotations/FSharpAnnotatedCacheProcessor.cs
```diff
@@ -0,0 +1,138 @@
...
+        var defaultMemberName = typeDeclaration.DeclaredName;
+        VisitParametersOwner(primaryConstructorDeclaration, defaultMemberName);
+
+        if (AttributeUtil.HasAttributeSuffix(defaultMemberName, out var nameWithoutAttributeSuffix))
+          VisitParametersOwner(primaryConstructorDeclaration, nameWithoutAttributeSuffix);
+      }
+
+      public override void VisitTopBinding(ITopBinding topBinding)
+      {
+        foreach (var declaration in topBinding.HeadPattern.Declarations)
```
auduchinok, line 74: `HeadPattern` may be `null`.

auduchinok (commented):
@DedSec256 Very nice work!

I've left some comments. Hope it helps :)

ReSharper.FSharp/src/FSharp.Psi.Features/src/Injected/FSharpLiteralInjectionTarget.fs
DedSec256: I think we can remove this code

ReSharper.FSharp/src/FSharp.Psi.Features/src/Injected/FSharpRegexProviders.fs
```diff
@@ -0,0 +1,58 @@
...
+type RegExprPsiProvider(injectorProvider: FSharpRegexInjectionProvider) =
+    inherit LiteralsInjectionPsiProvider<FSharpLanguage, ClrRegexLanguage>(injectorProvider, ClrRegexLanguage.Instance)
+
+    override _.ProvidedLanguageCanHaveNestedInjects = false
+
+
+[<Language(typeof<FSharpLanguage>)>]
+type FSharpRegularExpressionCompletionProvider() =
+    interface IRegexLanguageSpecificCompletionProvider with
+        override this.InitializeContext(owner, _) = owner :? ILiteralExpression
```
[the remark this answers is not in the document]
DedSec256: This code is invoked for nodes in which injections have already been checked. However, one really needs to add a check for ILiteralPat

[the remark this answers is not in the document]
DedSec256: This code is invoked only for nodes checked for injection earlier. This automatically works for all applicable FSharpString, including active patterns and type providers.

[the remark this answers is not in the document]
auduchinok: Can it then always return `true`? Why does it only check the expressions?

ReSharper.FSharp/src/FSharp.Psi/src/Annotations/FSharpAnnotatedCacheProcessor.cs
DedSec256: We also need to support probably the same case for ILocalBinding. And a simpler case with the regular local binding function.

DedSec256:
- [ ] FSharpAnnotatedMembersCacheProcessor (signatures/accessors, etc)

auduchinok (approved):
This is a really good work! Thanks @DedSec256 and @saul!

PR #136, opened 2020-06-09
Added more highlightings

En3Tho:
More highlightings for F#!

1. Computation expressions
2. Extension methods
3. Extension properties
~~4. Generic constraints (not a separate option, gets auto highlighted like method or property)~~
~~5. F# methods (let functions compiled as methods)~~ (goes to FSharp functions)
6. F# value functions
7. F# mutable value functions
8. Units of measure

Before:
image
After:
image

There is still work to do with parameter highlighting, but after spending some time and getting no result I've decided to try pulling this as it already covers about 90% of what I wanted to do

TODO:

- [x] Fix current tests. I need some advice here becase I have to change "gold" files and maybe you don't really need all those new highlighting options (althought I personanly find them all useful)
- [x] Add more tests
- [ ] Ask someone from compiler team about highlighting of method parameters (couldn't figure out how to do this one, sadly)


vasily-kirichenko:
String literals and properties are given almost the same color:

image

En3Tho:
Yes, but that's all up to the user. You can pick any color you want. I just wanted to demonstrate the ability to pick a color if you want extension methods/props to be more distinguishable

ReSharper.FSharp/src/Daemon.FSharp/src/FSharpSymbolHighlightingUtil.cs
```diff
@@ -51,22 +54,35 @@ public static string GetMfvHighlightingAttributeId([NotNull] this FSharpMemberOr
...
       if (mfv.IsActivePattern)
         return FSharpHighlightingAttributeIdsModule.ActivePatternCase;
 
-      if (mfv.IsMutable || mfv.IsRefCell())
-        return FSharpHighlightingAttributeIdsModule.MutableValue;
-
       if (IsMangledOpName(mfv.LogicalName))
         return FSharpHighlightingAttributeIdsModule.Operator;
 
+      if (mfv.IsValCompiledAsMethod)
```
auduchinok, line 75: This property is used to distinguish compiled representations of module values, it's not about how F# sees these symbols, I think it should not be used to highlight F# code.
We use it for looking at how F# symbols represented when used from other languages, but never inside F#.

ReSharper.FSharp/src/Daemon.FSharp/src/FSharpSymbolHighlightingUtil.cs
```diff
@@ -82,6 +98,13 @@ public static string GetHighlightingAttributeId([NotNull] this FSharpSymbol symb
         case FSharpEntity entity when !entity.IsUnresolved:
           return GetEntityHighlightingAttributeId(entity);
 
+        case FSharpParameter parameter: // FSharpParameter is a static member in a generic constraint, not a method parameter
+          if (parameter.FullName.Equals("new"))
+            return FSharpHighlightingAttributeIdsModule.Keyword;
```
auduchinok, line 103: Please add tests for this behaviour.

ReSharper.FSharp/src/Daemon.FSharp/src/FSharpSymbolHighlightingUtil.cs
```diff
@@ -82,6 +98,13 @@ public static string GetHighlightingAttributeId([NotNull] this FSharpSymbol symb
...
+          return parameter.Type.GenericArguments.Count == 2 && parameter.Type.GenericArguments[0].IsUnit
```
auduchinok, line 104: And the same here. It'd be much easier to reason about this code in future if it well-covered with tests.

ReSharper.FSharp/src/Daemon.FSharp/src/FSharpSymbolHighlightingUtil.cs
```diff
@@ -82,6 +98,13 @@ public static string GetHighlightingAttributeId([NotNull] this FSharpSymbol symb
         case FSharpEntity entity when !entity.IsUnresolved:
           return GetEntityHighlightingAttributeId(entity);
 
+        case FSharpParameter parameter: // FSharpParameter is a static member in a generic constraint, not a method parameter
```
auduchinok, line 101: AFAIK `FSharpParameter` is currently used for named args in method invocations and in type providers instantiations. Function parameters are represented as values reported as `FSharpMemberOrFunctionOrValue` symbols and I'm not sure if there's a way to tell them apart easily via `FSharpSymbol` APIs. Maybe it could be added the same way we've added `IsValCompiledAsMethod` previously.

ReSharper.FSharp/src/FSharp.Common/src/Util/FSharpHighlightingAttributeIds.fs
```diff
@@ -98,6 +103,22 @@ type FSharpSettingsNamesProvider() =
       Layer = HighlighterLayer.SYNTAX,
       EffectType = EffectType.TEXT, ForegroundColor = "#000000", DarkForegroundColor = "#B5CEA8");
 
+  RegisterHighlighter(
+      FSharpHighlightingAttributeIds.Measure,
+      FallbackAttributeId = FSharpHighlightingAttributeIds.TypeParameter,
+      GroupId = FSharpHighlightingAttributeIds.GroupId,
+      RiderPresentableName = "Syntax//Measure",
```
auduchinok, line 110: It should probably go to the `Types` group and to have `Unit of measure` name that is used in F# docs.

ReSharper.FSharp/src/FSharp.Common/src/Util/FSharpHighlightingAttributeIds.fs
```diff
@@ -98,6 +103,22 @@ type FSharpSettingsNamesProvider() =
...
+      Layer = HighlighterLayer.SYNTAX,      
+      EffectType = EffectType.TEXT, ForegroundColor = "DarkBlue", DarkForegroundColor = "LightBlue");
+  
+  RegisterHighlighter(
+      FSharpHighlightingAttributeIds.ComputationExpression,
+      FallbackAttributeId = FSharpHighlightingAttributeIds.Keyword,
+      GroupId = FSharpHighlightingAttributeIds.GroupId,
+      RiderPresentableName = "Syntax//Computation expression",
```
auduchinok, line 118: Please move it to `Values` group.

auduchinok:
> Fix current tests. I need some advice here becase I have to change "gold" files and maybe you don't really need all those new highlighting options (althought I personanly find them all useful)

It's OK to just update them with newer ids.

> Add more tests (if needed)

Please do. 🙂

ReSharper.FSharp/src/Daemon.FSharp/src/FSharpSymbolHighlightingUtil.cs
En3Tho: Well, my idea was to group methods (like C# methods) and mvf's with IsValCompiledAsMethod together and also have a clear difference between Methods and FSharpFuncs<_>. I thought this bool meant "F# compiler compiles this function as an actual method, not as FSharpFunc<_>. It happens when a function is bound to top level of module or class. Should I recheck this with csharplab or any other IL-viewer?
I had a different highlighting in at first for this kind of methods, but then it seemed kinda misleading and overlapped with a "Method" highlighting.
Do you propose trying to find another way of distinguishing methods and funcs or just group all fsharp funcs together not matter what they compile to?

ReSharper.FSharp/src/Daemon.FSharp/src/FSharpSymbolHighlightingUtil.cs
En3Tho: Thank you. This is important info. I guess I will have to add lots of different tests.. :)

ReSharper.FSharp/src/Daemon.FSharp/src/FSharpSymbolHighlightingUtil.cs
auduchinok: > I thought this bool meant "F# compiler compiles this function as an actual method, not as FSharpFunc<>. It happens when a function is bound to top level of module or class.

Yes, it's true for module values, though, not for classes.

> Do you propose trying to find another way of distinguishing methods and funcs or just group all fsharp funcs together not matter what they compile to?
I think I'd prefer it highlighted based on how it's seen inside F# code, not the compiled representation.

auduchinok: JFYI: fixed the wording about `IsValCompiledAsMethod`, it should be more clear now.

ReSharper.FSharp/src/Daemon.FSharp/src/FSharpSymbolHighlightingUtil.cs
```diff
@@ -82,6 +98,13 @@ public static string GetHighlightingAttributeId([NotNull] this FSharpSymbol symb
...
+          if (parameter.FullName.Equals("new"))
```
vasily-kirichenko, line 102: Why are you using `.Equals` instead of just `=`? Less noise for my taste.

ReSharper.FSharp/src/Daemon.FSharp/src/FSharpSymbolHighlightingUtil.cs
En3Tho: I see your point. I will remove that check. I guess I have 3 choices now as we won't really show the difference between methods and funcs (everything will be shown as methods).
1. Remove "FSharp Value Function Identifier", use "Method" instead. Remove "FSharp Mutable Value Function" and use "Mutable value". Will lose the ability to separate mutable funcs and mutable values but maybe it's not needed that much?
2. Rename to just "FSharp Function" and "FSharp mutable function" or something to else?
3. Leave them as they are.
Now that we won't make distinction between Methods and Funcs they actually seem kinda redundant.
I will go with whatever decision you make.

auduchinok: > we won't really show the difference between methods and funcs (everything will be shown as methods).

It looks it's covered by IsFunctionType check below.

I think just removing this check will make these corner-case-ish values continue to be highlighted as values (which currently seems fine to me), and the symbols that seen as functions should not be affected at all (due to being covered separately).

En3Tho: Yeah. But I just want to know what to do with those identifiers I've added. Do we need them at all?

auduchinok: Yes, they all seem to be good, thanks!

I'd probably prefer ValueFunction renamed to Function, but it's subjective and is not that important. 🙂

ReSharper.FSharp/src/Daemon.FSharp/src/FSharpSymbolHighlightingUtil.cs
En3Tho: This code was removed.

ReSharper.FSharp/src/Daemon.FSharp/src/FSharpSymbolHighlightingUtil.cs
En3Tho: This code was removed.

ReSharper.FSharp/src/FSharp.Common/src/Util/FSharpHighlightingAttributeIds.fs
En3Tho: Fixed

ReSharper.FSharp/src/Daemon.FSharp/src/FSharpSymbolHighlightingUtil.cs
En3Tho: This code was removed.

ReSharper.FSharp/src/FSharp.Common/src/Util/FSharpHighlightingAttributeIds.fs
En3Tho: Fixed.

ReSharper.FSharp/src/Daemon.FSharp/src/FSharpSymbolHighlightingUtil.cs
En3Tho: Thanks. But this code was removed

En3Tho:
Sorry for the delay. I've fixed old tests and added new ones. Please tell me if there is more work I need to do on this thing.
Sadly, I couldn't find a way to get this FSharpParameter thing with static constrains to work properly. F# compiler knows that `when 'T : (static member Two : 'T)` - this is a Property and `when 'T : (static member Two : unit -> 'T)` - now this is a Method but it's either me or API does not give out this kind of info that easily
So it was really easy to confuse properties and methods. Also it gave false results at object instantiation in case like this: 
image
I guess it's better to just leave those things highlighted as simple values for now.
Also, I will note again that FSharp Method's and FSharp Function's are not the same thing - FSharp Method is something we get from the "outside" (e.g. C#) and FSharp Function comes from F#. F# function can either be method or func but we have decided to not give out this kind of info via highlighting.
Also, should I check for [<Extension>] Attribute when highlighting extension methods? It's mainly for C# I guess and F# shouldn't really care about it internally. But on the other side it might be useful.

ReSharper.FSharp/src/Daemon.FSharp/src/FSharpSymbolHighlightingUtil.cs
En3Tho: Renamed to Function.

En3Tho:
Also, without IsValCompiledAsMethodCheck things like these are not being shown as methods for some reason: 
image
vs
image
I find this a kind of bug at F# comp side because it's a value, it's compiled as method, but for some reason it's not a function type.

ReSharper.FSharp/test/data/features/daemon/identifierHighlighting/Functions 02.fs.gold
```diff
@@ -0,0 +1,115 @@
...
+    for |i|(53) = 0 to |arr|(54).|Length|(55) do
+        |arr|(56).[|i|(57)] <- |typeof|(58)<|'T|(59)>
+    let |element|(60) = (|^T|(61) : (static member |Two|(62) : |unit|(63) -> |'T|(64)) ())
+    let |arr2|(65) = Enumerable.Repeat(element, 10) ||>|(66) |Seq|(67).|toArray|(68)
+    |arr|(69) ||>|(70) |Array|(71).|zip|(72) |arr2|(73)
+    
+let inline |arraysAndStaticProp|(74)<|'T|(75) when |'T|(76) : (static member |Two|(77) : |'T|(78))> =
+    (|^T|(79) : (static member |Two|(80) : |'T|(81)) ()) ||>|(82) |Seq|(83).|replicate|(84) 10
+    ||>|(85) |Seq|(86).|toArray|(87)
+    ||>|(88) |Array|(89).|zip|(90) ((Enumerable.Repeat(typedefof<'T>, 10)).ToArray())
```
En3Tho, line 23: `((Enumerable.Repeat(typedefof<'T>, 10)).ToArray())` - for some reason auto generation did not work for this thing.
Should I manually fix this?

En3Tho: There are more tests with similar bugs. Linq test is totally broken. Just realized this.

auduchinok: Does adding `[<TestReferences("System", "System.Core")>]` to the test case or to `IdentifierHighlightingTest` class help? (Maybe a single reference instead of both is enough.)

auduchinok: Could you please also try splitting the bigger tests here into smaller ones so it's easier to see particular things being tested?

ReSharper.FSharp/test/data/features/daemon/identifierHighlighting/Computation expressions.fs
```diff
@@ -0,0 +1,7 @@
+open System
+open System.Threading.Tasks
+
+let asyncCex = async { let! a = Task.Run(fun x -> 10) |> Async.AwaitTask
+                       a |> ignore } |> Async.RunSynchronously
```
auduchinok, line 5: Could it be something simple, like the following?
```fsharp
let asyncCex = async { return 1 }
```

En3Tho: Will do

ReSharper.FSharp/test/data/features/daemon/identifierHighlighting/Extension members 02.fs.gold
```diff
@@ -0,0 +1,22 @@
+﻿open |System|(0)
+open |System|(1).Linq
+
+type |Char|(2) with
+    static member |IsZero|(3) |x|(4) = |x|(5) |=|(6) |char|(7) 0
+
+let |linq|(8) (|x|(9) : |string|(10)) =
+    |x|(11).Select(Char.IsZero).ToArray()
```
En3Tho, line 8: How can I make this thing work? It skips Select and ToArray. Maybe add more references somewhere?

ReSharper.FSharp/test/data/features/daemon/identifierHighlighting/Functions 02.fs.gold
En3Tho: It helped. Thank you!

auduchinok:
@En3Tho Thanks for this change!
I'm glad you've found a way to highlight the type functions too (I wasn't sure if it's possible to see it via FCS right now).

auduchinok:
Sorry, I've messed it up a bit trying to push a small cleanup. The changes got merged without closing the PR automatically, so I'll close it manually.

En3Tho:
No problem. Thank you for accepting :)

PR #148, opened 2020-07-20
Make RemoveNeverMatchingRuleFix scoped

reacheight:
never_matching_rule2


vasily-kirichenko:
It'd be more natural to have "Remove never matching ruleS", your example makes this particularly clear.

PR #151, opened 2020-07-20
Add parens to application QuickFix for NotAFunction error

DedSec256:
https://github.com/JetBrains/fsharp-support/pull/142

Depends on:
https://github.com/auduchinok/visualfsharp/pull/3
https://github.com/auduchinok/visualfsharp/pull/6


auduchinok:
I've made a small adjustment: when there's a single candidate and there's a way to present it, we'll show a bit better text for the quick fix:
Screenshot 2020-07-28 at 17 32 12

auduchinok:
@DedSec256 Thanks for this quick fix, it's a really cool thing to use! 🙂

PR #168, opened 2020-08-15
Fix spaces adding in RemoveRedundantParentExprFix

DedSec256:
From 
```
Add(5) -> Add5
```

To
```
Add(5) -> Add 5
```

### DIFF
--- a/ReSharper.FSharp/src/FSharp.Psi.Features/src/QuickFixes/RemoveRedundantParenExprFix.fs
+++ b/ReSharper.FSharp/src/FSharp.Psi.Features/src/QuickFixes/RemoveRedundantParenExprFix.fs
@@ -3,6 +3,7 @@
 open JetBrains.ReSharper.Plugins.FSharp.Psi
 open JetBrains.ReSharper.Plugins.FSharp.Psi.Features.Daemon.Highlightings
 open JetBrains.ReSharper.Plugins.FSharp.Psi.Features.Daemon.QuickFixes
+open JetBrains.ReSharper.Plugins.FSharp.Psi.Impl.Tree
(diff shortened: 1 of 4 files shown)


ReSharper.FSharp/src/FSharp.Psi.Features/src/QuickFixes/RemoveRedundantParenExprFix.fs
```diff
@@ -27,5 +29,15 @@ type RemoveRedundantParenExprFix(warning: RedundantParenExprWarning) =
         let innerExprIndent = innerExpr.Indent
         let indentDiff = parenExprIndent - innerExprIndent
 
+        let parent = parenExpr.Parent
+
+        if not (parent :? IParenExpr) then 
+            let prevSibling = parenExpr.PrevSibling
+            if isNotNull prevSibling && not (prevSibling.IsWhitespaceToken()) then 
+                ModificationUtil.AddChildBefore(parenExpr, Whitespace()) |> ignore
+
```
reacheight, line 38: We should also check the paren expr next sibling:
```f#
let inc x = x + 1
let a = (inc)3
```

ReSharper.FSharp/test/data/features/quickFixes/removeRedundantParens/expr/App 03 - Add spaces.fs
```diff
@@ -0,0 +1,4 @@
+module Module
+
+let add a b c = a + b + c 
+let x = 1({caret}2)3
```
reacheight, line 4: Is `add` supposed to be here?

PR #179, opened 2020-09-15
Add syntax highlighted tooltips

saul:
image

### DIFF
--- a/ReSharper.FSharp/src/FSharp.Psi.Features/src/Daemon/Stages/FSharpIdentifierTooltipProvider.fs
+++ b/ReSharper.FSharp/src/FSharp.Psi.Features/src/Daemon/Stages/FSharpIdentifierTooltipProvider.fs
@@ -1,7 +1,6 @@
 module JetBrains.ReSharper.Plugins.FSharp.Daemon.Stages.Tooltips
 
 open System
-open System.Collections.Generic
 open FSharp.Compiler.Layout
 open FSharp.Compiler.SourceCodeServices
 open JetBrains.DocumentModel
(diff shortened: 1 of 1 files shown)


drhumlen:
Looks really nice, which helps a lot on the readability. 

Would be nice to have this merged asap so I can statt using it 😃

What does it look like when you hover other things? like local on regular functions or a record?

ReSharper.FSharp/src/FSharp.Psi.Features/src/Daemon/Stages/FSharpIdentifierTooltipProvider.fs
```diff
@@ -43,61 +107,59 @@ type FSharpIdentifierTooltipProvider(lifetime, solution, presenter, xmlDocServic
...
                       | _ -> () ]
-                    |> String.concat "\n\n"
-                    |> result.Add))
-
-        result.Join(RiderTooltipSeparator)
-
-    override x.GetRichTooltip(highlighter) =
-        RichTextBlock(x.GetTooltip(highlighter))
+                    |> richTextJoin "\n\n"))
+        // todo: add support for _RIDER_HORIZONTAL_LINE_TOOLTIP_SEPARATOR_
```
auduchinok, line 161: @kirillgla Is it supposed to work with the rich text tooltips?

kirillgla: Yes, it is. I tried it out locally, and I can say that replacing `"\n--------------------\n"` with `"_RIDER_HORIZONTAL_LINE_TOOLTIP_SEPARATOR_"` works just like one would expect it to
image

kirillgla:
Also, I strongly recommend implementing `IRiderQuickDocTooltipRequest` in F#. While syntax highlighting in tooltips might work without it, implementing that interface will make the IDE UI more consistent

auduchinok:
@kirillgla Thanks for the advice! We'll look into migrating to these APIs.

ReSharper.FSharp/src/FSharp.Psi.Features/src/Daemon/Stages/FSharpIdentifierTooltipProvider.fs
auduchinok: Thanks! 🙂

auduchinok (approved):

auduchinok:
@saul Thanks again for this, it looks really good now!
And props to @kirillgla for the rich text tooltips in the platform. 🙂

PR #186, opened 2020-10-05
Tooltip improvements

saul:
image

Changes:

- Fixed 'UnknownEntity' being highlighted as a type rather than a value
- Spaces at the start of tooltip lines are no longer ignored


auduchinok (approved):
@saul Thanks!

PR #191, opened 2020-10-18
Add CI for ReSharper.FSharp (Win only)

DedSec256:
PR check examples: 
- failed build: https://github.com/DedSec256/fsharp-support/pull/4
- failed tests:  https://github.com/DedSec256/fsharp-support/pull/3

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,55 @@
+name: "build-test"
+on:
+  push:
+    branches:
+      - ber.a/ci
+  pull_request:
+    branches:
+      - ber.a/ci
+
+jobs:
+  test-backend:
+    runs-on: ${{ matrix.image }}
+    strategy:
+      matrix:
+        image: [windows-latest]
+    env:
+      NUGET_PACKAGES: ${{ github.workspace }}/.github/nuget-packages
+      GRADLE_USER_HOME: ${{ github.workspace }}/.github/gradle
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Gradle Wrapper Cache
+        uses: actions/cache@v1.1.0
+        with:
+          path: ${{ env.GRADLE_USER_HOME }}/wrapper
+          key: ${{ runner.os }}.gradle-wrapper.${{ hashFiles('rider-fsharp/gradle/**') }}
+
+      - name: Gradle Wrapper Validation
+        uses: gradle/wrapper-validation-action@v1.0.3
+
+      - name: NuGet Cache
+        uses: actions/cache@v1.1.0
+        with:
+          path: ${{ env.NUGET_PACKAGES }}
+          key: ${{ runner.os }}.nuget.${{ hashFiles('ReSharper.FSharp/**/*proj') }}+${{ hashFiles('ReSharper.FSharp/**/*.props') }}
+
+      - name: Gradlew Prepare
+        working-directory: ./rider-fsharp
+        run: ./gradlew prepare
+
+      - name: Add MSBuild to PATH # for tests
+        uses: microsoft/setup-msbuild@v1.0.2
+
+      - name: Build ReSharper.FSharp
+        id: build
+        working-directory: ./ReSharper.FSharp
+        run: dotnet build
+
+      - name: Test ReSharper.FSharp
+        working-directory: ./ReSharper.FSharp
+        run: dotnet test --no-build
+
+      - name: Stop Gradle Daemon # to collect Gradle cache
+        working-directory: ./rider-fsharp
+        run: ./gradlew --stop
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,6 +50,10 @@ jobs:
         working-directory: ./ReSharper.FSharp
         run: dotnet test --no-build
 
+      - name: Build Plugin Frontend
+        working-directory: ./rider-fsharp
+        run: gradlew buildPlugin
+
       - name: Stop Gradle Daemon # to collect Gradle cache
         working-directory: ./rider-fsharp
         run: ./gradlew --stop
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,11 +52,11 @@ jobs:
 
       - name: Build Plugin Frontend
         working-directory: ./rider-fsharp
-        run: gradlew buildPlugin
+        run: ./gradlew buildPlugin
 
       - name: Test BuildSearchableOptions
         working-directory: ./rider-fsharp
-        run: gradlew buildSearchableOptions
+        run: ./gradlew buildSearchableOptions
 
       - name: Stop Gradle Daemon # to collect Gradle cache
         working-directory: ./rider-fsharp
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,7 +42,6 @@ jobs:
         uses: microsoft/setup-msbuild@v1.0.2
 
       - name: Build ReSharper.FSharp
-        id: build
         working-directory: ./ReSharper.FSharp
         run: dotnet build
 
@@ -52,11 +51,8 @@ jobs:
 
       - name: Build Plugin Frontend
         working-directory: ./rider-fsharp
+        if: ${{ success() }} || ${{ failure() }}
         run: ./gradlew buildPlugin
-
-      - name: Test BuildSearchableOptions
-        working-directory: ./rider-fsharp
-        run: ./gradlew buildSearchableOptions
 
       - name: Stop Gradle Daemon # to collect Gradle cache
         working-directory: ./rider-fsharp
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,11 +1,9 @@
 name: "build-test"
 on:
   push:
-    branches:
-      - ber.a/ci
+    branches: [ $default-branch ]
   pull_request:
-    branches:
-      - ber.a/ci
+    branches: [ $default-branch ]
 
 jobs:
   test-backend:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,6 +54,10 @@ jobs:
         working-directory: ./rider-fsharp
         run: gradlew buildPlugin
 
+      - name: Test BuildSearchableOptions
+        working-directory: ./rider-fsharp
+        run: gradlew buildSearchableOptions
+
       - name: Stop Gradle Daemon # to collect Gradle cache
         working-directory: ./rider-fsharp
         run: ./gradlew --stop
--- a/ReSharper.FSharp/test/src/FSharp.Common.Tests/FSharp.Common.Tests.fsproj
+++ b/ReSharper.FSharp/test/src/FSharp.Common.Tests/FSharp.Common.Tests.fsproj
@@ -15,8 +15,11 @@
   <ItemGroup>
     <PackageReference Include="FSharp.Core" Version="$(FSharpCoreVersion)" />
     <PackageReference Include="JetBrains.FSharp.Compiler.Service" Version="$(FSharpCompilerServiceVersion)" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.7.0" />
     <PackageReference Include="Moq" Version="4.8.2" />
     <PackageReference Include="NUnit.Console" Version="3.10.0" />
+    <PackageReference Include="NUnit3TestAdapter" Version="3.17.0" />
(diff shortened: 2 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,73 @@
...
+        uses: actions/cache@v1.1.0
+        with:
+          path: ${{ env.GRADLE_USER_HOME }}/wrapper
+          key: ${{ runner.os }}.gradle-wrapper.${{ hashFiles('rider-fsharp/gradle/**') }}
+
+      - name: NuGet Cache
+        uses: actions/cache@v1.1.0
+        with:
+          path: ${{ env.NUGET_PACKAGES }}
+          key: ${{ runner.os }}.nuget.${{ hashFiles('ReSharper.FSharp/**/*.csproj') }}+${{ hashFiles('ReSharper.FSharp/**/*.props') }}
```
auduchinok, line 39: Is there a reason you're only interested in `csproj` projects and not in `fsproj` ones?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,73 @@
...
+
+      - name: Gradlew Prepare
+        working-directory: ./rider-fsharp
+        run: ./gradlew prepare
+
+      - name: Setup .NET Core SDK
+        uses: actions/setup-dotnet@v1.7.2
+        with:
+          dotnet-version: '3.1.101'
```
auduchinok, line 48: Is this the same version as on the internal CI? Is there a way we could use a single place to specify the version for both CIs?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,73 @@
...
+
+      - name: Add MSBuild to PATH
+        uses: microsoft/setup-msbuild@v1.0.2
```
auduchinok, line 51: Do we need to add MSBuild to the path? We don't use it for the plugin build itself.
Is it for the tests run? Could you add a comment if yes?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,73 @@
+name: "build-test"
+on:
+  push:
+    branches:
+      - net203
```
auduchinok, line 5: Is there a way to specify the default branch instead of hardcoding one? We've been changing the main branch for each release recently and it'd be great not to have to update it here separately every time.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,73 @@
...
+          path: build/gradle-jvm
+          key: ${{ runner.os }}.gradle-jvm.${{ hashFiles('rider-fsharp/gradle*') }}
+
+      - name: Gradle Wrapper Cache
+        uses: actions/cache@v1.1.0
+        with:
+          path: ${{ env.GRADLE_USER_HOME }}/wrapper
+          key: ${{ runner.os }}.gradle-wrapper.${{ hashFiles('rider-fsharp/gradle/**') }}
+
+      - name: NuGet Cache
```
auduchinok, line 35: Looking at the [example run](https://github.com/DedSec256/fsharp-support/pull/3/checks), it doesn't seem to work (if I understand the log correctly):
```
Cache not found for input keys: Windows.nuget.457abc6dcb0d1d36ebcee59dbef4d2a5abd8dae0d906beaab873caaf8d23ecf9+da217cfbad83434fb9208c23a2bf3aa7e58bdf8b88ba1c399320075ac49c0e8f.
```

ReSharper.FSharp/test/src/FSharp.Common.Tests/FSharp.Common.Tests.fsproj
```diff
@@ -15,8 +15,11 @@
   <ItemGroup>
     <PackageReference Include="FSharp.Core" Version="$(FSharpCoreVersion)" />
     <PackageReference Include="JetBrains.FSharp.Compiler.Service" Version="$(FSharpCompilerServiceVersion)" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.7.0" />
     <PackageReference Include="Moq" Version="4.8.2" />
     <PackageReference Include="NUnit.Console" Version="3.10.0" />
+    <PackageReference Include="NUnit3TestAdapter" Version="3.17.0" />
+    <PackageReference Include="NunitXml.TestLogger" Version="2.1.62" />
```
auduchinok, line 22: Please make sure it doesn't affect local test runs and internal CI.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,73 @@
...
+      - name: Setup .NET Core SDK
+        uses: actions/setup-dotnet@v1.7.2
+        with:
+          dotnet-version: '3.1.101'
+
+      - name: Add MSBuild to PATH
+        uses: microsoft/setup-msbuild@v1.0.2
+
+      # Build and Test ReSharper.FSharp
+      - name: Test FSharp.Common.Tests
```
auduchinok, line 54: Would it be better to have build and test as separate stages (so it's easier to see whether it's build or tests who's failing)?

auduchinok (commented):
It looks very promising. 🙂 
I left a few comments, please take a look.

Also the frontend tests run is missing.

.github/workflows/ci.yml
```diff
@@ -54,9 +54,9 @@ jobs:
       - name: Test FSharp.Common.Tests
         working-directory: ./ReSharper.FSharp
         run: dotnet test test/src/FSharp.Common.Tests/FSharp.Common.Tests.fsproj --logger:nunit --results-directory TestResults/FSharp.Common.Tests
+        continue-on-error: true
 
       - name: Test FSharp.Tests
-        if: always()
```
auduchinok: Don't we want to continue if these tests fail?

.github/workflows/ci.yml
DedSec256: yes this is for tests

.github/workflows/ci.yml
DedSec256: oh, really

.github/workflows/ci.yml
DedSec256: think we can do it with the ```$default-branch``` macro

.github/workflows/ci.yml
DedSec256: internal CI uses SDK v3.0.100

.github/workflows/ci.yml
DedSec256: When key doesn't match an existing cache, it's called a cache miss, and a new cache is created if the job completes successfully

.github/workflows/ci.yml
DedSec256: I think it will be much slower as ```dotnet test``` builds the project too

.github/workflows/ci.yml
DedSec256: now it works exactly like that

.github/workflows/ci.yml
auduchinok: Why would it be slower? Does it do a rebuild?

ReSharper.FSharp/test/src/FSharp.Common.Tests/FSharp.Common.Tests.fsproj
DedSec256: everything is green there (:

DedSec256: everything is green there (:

.github/workflows/ci.yml
ForNeVeR: For the record, [here's a sample of syntax](https://github.com/actions/starter-workflows/blob/08ff79c7b930315861d16bed4903ddd90bf7c05c/ci/dotnet-core.yml#L5) if you need one.

.github/workflows/ci.yml
ForNeVeR: Please check if the cache works at all.

In my experiments, the overall cache volume required for the Rider plugin was too big (more than 2 GiB), so the resulting cache was being destroyed momentarily after creation.

.github/workflows/ci.yml
ForNeVeR: First, if everything is set up correctly, pre-test build should be no-op.

Second, even if everything's not set up correctly, there's `dotnet test --no-build`.

ForNeVeR (approved):
Approved in principle, [as our glorious leader says](https://github.com/fsharp/fslang-suggestions/issues?q=is%3Aissue+is%3Aopen+label%3Aapproved-in-principle).

.github/workflows/ci.yml
saul: You can also `dotnet test` the solution file. That will test all projects in the solution, rather than you having to specify each test project. (It may also test the assemblies in parallel - not sure.)

.github/workflows/ci.yml
saul: Note this SDK is already preinstalled on the agent. See https://github.com/actions/virtual-environments/blob/main/images/win/Windows2019-Readme.md

.github/workflows/ci.yml
citizenmatt: AIUI, this means run on all pushes to `net203` and all pull requests targeting `net203`. This is effectively the same as "everything", so why not just specify a default `on: [push, pull_request]` or just `on: [pull_request]`? 

This would give you CI results for pushes to an active PR, and simplifies config.

.github/workflows/ci.yml
citizenmatt: I tried setting up cache keys for the Unity plugin ([JetBrains/resharper-unity#1683](https://github.com/JetBrains/resharper-unity/pull/1683)) and eventually closed it without merging, because it didn't make any difference to performance - the workflow was essentially downloading zip files from a cache, or zip files from maven/gradle/whatever. So added configuration for no performance gain.

ForNeVeR: For Azure DevOps IntelliJ (where we're unable to use build system caches due to Azure being bad), there's a very big difference.

In my experience, about 30% of all build failures are from Gradle not being able to download something (i.e. due to network errors). Utilizing the caches would nullify this kind of errors, and improve the overall CI success rate, which couldn't be a bad thing.

So, I would recommend to use cache if it's possible.

citizenmatt:
It would probably be a good idea to add a [Gradle wrapper validation](https://github.com/marketplace/actions/gradle-wrapper-validation) step. This will make sure that any updates to `gradle-wrapper.jar` are validated against known checksums from the Gradle project, and you know you're not running malicious code (of course, there may be malicious code added to tests, etc. but this is a simple easy check to add)

.github/workflows/ci.yml
auduchinok: This would also split build warnings from test errors, which both are in the same task now, making finding test errors rather difficult: https://github.com/DedSec256/fsharp-support/pull/3/checks?check_run_id=1272374074

.github/workflows/ci.yml
auduchinok: >  a new cache is created if the job completes successfully
OK, I see, thanks. For some reason I expected it to create the cache anyway.

.github/workflows/ci.yml
auduchinok: Why `continue-on-error: true` is added to `Test FSharp.Common.Tests` but not to `Test FSharp.Tests`? Does it affect both configurations somehow? Will next configurations (e.g. frontend tests) run properly if `Test FSharp.Tests` fails?

.github/workflows/ci.yml
auduchinok: @citizenmatt Wouldn't it also run on every push to other branches? I'm not sure if it's a good idea.

ForNeVeR: I could now remember why I always prefer to specify the branch here.

For `on: [push, pull_request]` it will build some of the PRs _twice_, since you push them to this repository, and then create pull requests with them.

If you have three platforms, it becomes messy very quickly:
image

So, let's go with `$default-branch`, and not just `on: [push, pull_request]`.

auduchinok:
Looking at the example runs, it shows a step as successful, despite it having some failing tests. Please take a look.
Screenshot 2020-10-26 at 13 14 36

.github/workflows/ci.yml
auduchinok: @DedSec256 Have you had a chance to make sure this cache is working as expected?

DedSec256: @auduchinok, yes (:
image
But I removed the cache for the Gradle JVM since it worked strangely

auduchinok (approved):
@DedSec256 LGTM, thanks! 🙂

PR #206, opened 2020-11-19
Generate idea-version during build

citizenmatt:
The `plugin.xml` file currently doesn't include `<idea-version … />` which is required for using the F# plugin as a dependency (e.g. to add F# support to another Rider plugin). If it is missing, the development version of the IDE used to build the plugin will fail to load the F# plugin.

Resetting the `updateSinceUntilBuild` back to its default value of `true` will add the element into the copy of the file used during build. The gradle scripts will set the current value of the Rider IJ platform as the `since-build` and will use a wildcard for the `until-build`, e.g.:

```xml
<idea-version since-build="203.5949" until-build="203.*" />
```

### DIFF
--- a/rider-fsharp/build.gradle.kts
+++ b/rider-fsharp/build.gradle.kts
@@ -75,7 +75,6 @@ intellij {
 
     instrumentCode = false
     downloadSources = false
-    updateSinceUntilBuild = false
 
     // Workaround for https://youtrack.jetbrains.com/issue/IDEA-179607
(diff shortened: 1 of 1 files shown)


auduchinok (approved):
Oh, not setting it was helpful in sharing a plugin build with someone using Rider dev builds (that could contain no plugin bundled), so it'd allow using the plugin in an older build. It was some time ago and I haven't needed it recently, though, so I'm totally fine with it.

PR #218, opened 2020-12-28
Add ConvertTupleToArrayOrListElementsFix

DedSec256:
JetBrains


ReSharper.FSharp/src/FSharp.Psi.Features/src/Daemon/Stages/FcsErrorsStageProcessBase.fs
```diff
@@ -68,6 +69,17 @@ module FSharpErrors =
     let [<Literal>] ifExprMissingElseBranch = "This 'if' expression is missing an 'else' branch."
     let [<Literal>] expressionIsAFunctionMessage = "This expression is a function value, i.e. is missing arguments. Its type is "
 
+    let private unexpectedTypeError1 = Regex("This expression (was) expected to have type\n    '(.+)'    \nbut here has type\n    '(.+)'")
+    let private unexpectedTypeError2 = Regex("All elements of (an array|a list) must be of the same type as the first element, which here is '(.+)'. This element has type '(.+)'.")
+
+    let tryGetUnexpectedTypeErrorInfo message =
```
auduchinok, line 75: Could we try using an active pattern here? :)
http://www.fssnip.net/29/title/Regular-expression-active-pattern

ReSharper.FSharp/src/FSharp.Psi.Features/src/Daemon/Stages/FcsErrorsStageProcessBase.fs
```diff
@@ -68,6 +69,17 @@ module FSharpErrors =
...
+        let unexpectedTypeMatch = unexpectedTypeError1.Match(message)
+        if unexpectedTypeMatch.Success then unexpectedTypeMatch.Groups else
+
+        let unexpectedTypeMatch = unexpectedTypeError2.Match(message)
```
auduchinok, line 79: It likely worths adding a separate generated highlighting for this error. Are they separate in FCS?

ReSharper.FSharp/src/FSharp.Psi.Features/src/QuickFixes/ConvertTupleToElementsFix.fs
```diff
@@ -0,0 +1,76 @@
...
+
+    let actual = warning.ActualType
+    let expr = warning.Expr
+    let parent = ConvertTupleToElements.getArrayOrListParent expr
+
+    override x.Text =
+        if parent.IsArray then "Convert tuple to array elements" else "Convert tuple to list elements"
+
+    override x.IsAvailable _ = ConvertTupleToElements.isAvailable expr actual
+    override x.ExecutePsiTransaction _ =
```
auduchinok, line 52: New line is missing before the multi-line member.

ReSharper.FSharp/src/FSharp.Psi/src/Impl/Tree/ArrayOrListExpr.cs
```diff
@@ -0,0 +1,10 @@
+﻿using JetBrains.ReSharper.Plugins.FSharp.Psi.Parsing;
+using JetBrains.ReSharper.Psi.Tree;
+
+namespace JetBrains.ReSharper.Plugins.FSharp.Psi.Impl.Tree
+{
+  internal partial class ArrayOrListExpr
+  {
+    public bool IsArray => FirstChild?.GetTokenType() == FSharpTokenType.LBRACK_BAR;
```
auduchinok, line 8: It'd probably be better to have separate types for lists and arrays in the psi grammar. The current design comes from FCS.

ReSharper.FSharp/test/data/features/quickFixes/convertTupleToElements/Large tuple.fs
```diff
@@ -0,0 +1,3 @@
+let f (x: int array) = ()
+
+f [|1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22{caret}|]
```
auduchinok, line 3: Any reason this particular length was chosen? Does it add something over the smaller tests?

ReSharper.FSharp/src/FSharp.Psi.Features/src/QuickFixes/ConvertTupleToElementsFix.fs
```diff
@@ -0,0 +1,76 @@
...
+                if isNotNull seq then ArrayOrListExprNavigator.GetByExpression(seq) else null
+        arrayOrList
+
+    let isAvailable expr actual =
+        isValid expr && isNotNull (getArrayOrListParent expr) && regex.Match(actual).Value = actual
+
+    let convertTupleToElements (expr: IFSharpExpression) (factory: IFSharpElementFactory) =
+        match expr with
+        | :? ITupleExpr as tuple ->
+            let sequentialExpr = factory.CreateSequentialExpr(tuple.ExpressionsEnumerable)
```
auduchinok, line 35: Could you check this approach on samples like:

Split over multiple lines:
```fsharp
[ 1, 2,
  3, 4 ]
```

Comments:
```fsharp
[ 1,

  // a very important comment here
  2 ]
```

I'd try creating the sequence expression via `ElementType.SEQUENTIAL_EXPR.Create()` and then moving the children nodes there, if needed.

Since the quick fix is also available in cases where the tuple is not the only list element (`[1, 2; 3]`), you have to make sure you don't get nested sequence expressions. Speaking about it, I'm not yet convinced we need this variant of the quick fix (at least now, when types info we have is somewhat limited here), since the main idea was about fixing wrong separators for the whole lists and not about fixing inner expressions in lists.

ReSharper.FSharp/src/FSharp.Psi.Features/src/QuickFixes/ConvertTupleToElementsFix.fs
```diff
@@ -0,0 +1,76 @@
...
+
+type ConvertTupleToElementsFix(warning: UnexpectedExpressionTypeError) =
+    inherit FSharpQuickFixBase()
+
+    let actual = warning.ActualType
+    let expr = warning.Expr
+    let parent = ConvertTupleToElements.getArrayOrListParent expr
+
+    override x.Text =
+        if parent.IsArray then "Convert tuple to array elements" else "Convert tuple to list elements"
```
auduchinok, line 49: `parent` sounds like `ITreeNode.Parent: ITreeNode` and _apparently_ it has a different type. Could it be something like `arrayOrListExpr` instead, giving a hint about its type?

ReSharper.FSharp/test/data/features/quickFixes/convertTupleToElements/Large tuple.fs
DedSec256: just checking that the regex recognizes looped variable names
image

ReSharper.FSharp/src/FSharp.Psi.Features/src/Daemon/Highlightings/FcsErrors.xml
```diff
@@ -299,4 +299,17 @@
     <QuickFix>NamespaceToModuleFix</QuickFix>
   </Error>
 
+  <Error staticGroup="FSharpErrors" name="UnexpectedExpressionType">
```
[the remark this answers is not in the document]
DedSec256: No, in FCS this is called *ErrorFromAddingTypeEquation1*

ReSharper.FSharp/test/data/features/quickFixes/convertTupleToElements/Large tuple.fs
auduchinok: Cool, thanks.

ReSharper.FSharp/src/FSharp.Psi.Features/src/Daemon/Highlightings/FcsErrors.xml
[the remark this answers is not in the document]
auduchinok: :(

ReSharper.FSharp/src/FSharp.Psi.Features/src/QuickFixes/ConvertTupleToElementsFix.fs
DedSec256: forgot to rename before pushing, thanks

ReSharper.FSharp/src/FSharp.Psi.Features/src/Daemon/Stages/FcsErrorsStageProcessBase.fs
DedSec256: Yes, they are

ReSharper.FSharp/src/FSharp.Psi.Features/src/Daemon/Highlightings/FcsErrors.xml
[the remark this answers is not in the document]
auduchinok: Would you like something like `AddTypeEquation`, so it's different from other unexpected type errors?

ReSharper.FSharp/src/FSharp.Psi.Features/src/QuickFixes/ConvertTupleToArrayOrListElementsFix.fs
```diff
@@ -0,0 +1,43 @@
...
+open JetBrains.ReSharper.Resources.Shell
+
+type ConvertTupleToArrayOrListElementsFix(expr: IFSharpExpression, actualType: string) =
+    inherit FSharpQuickFixBase()
+
+    let regex = Regex("'\w\d*( \* '\w\d*)+")
+    let arrayOrListExpr = ArrayOrListExprNavigator.GetByExpression(expr.IgnoreParentParens())
+    let isArray = arrayOrListExpr :? IArrayExpr
+
+    new (warning: AddTypeEquationError) =
```
auduchinok, line 22: Any reason the second constructor is needed?

ReSharper.FSharp/src/FSharp.Psi.Features/src/QuickFixes/ConvertTupleToArrayOrListElementsFix.fs
```diff
@@ -0,0 +1,43 @@
...
+        ConvertTupleToArrayOrListElementsFix(warning.Expr, warning.ActualType)
+
+    override x.Text =
+        if isArray then "Convert tuple to array elements" else "Convert tuple to list elements"
```
auduchinok, line 26: Just an idea: the reported error is about using a wrong type, and we guess it happens due to a wrong separator use. For a newcomer it may be not that obvious why the existing code is not a few list/array elements (in fact, the tuple _is_ such an element!) or why do we want to convert it. What if the text was something like `"Use ';' separators"`? Would it be easier to grasp what went wrong and why it would fix the code?

ReSharper.FSharp/src/FSharp.Psi.Features/src/QuickFixes/ConvertTupleToArrayOrListElementsFix.fs
```diff
@@ -0,0 +1,43 @@
...
+        isValid expr && isNotNull arrayOrListExpr && regex.Match(actualType).Value = actualType
+
+    override x.ExecutePsiTransaction _ =
+        use writeCookie = WriteLockCookie.Create(expr.IsPhysical())
+        use disableFormatter = new DisableCodeFormatter()
+
+        match expr with
+        | :? ITupleExpr as tuple ->
+            for comma in tuple.Commas do
+                replace comma (FSharpTokenType.SEMICOLON.CreateLeafElement())
```
auduchinok, line 38:
```suggestion
                replaceWithToken comma FSharpTokenType.SEMICOLON
```

ReSharper.FSharp/src/FSharp.Psi.Features/src/QuickFixes/ConvertTupleToArrayOrListElementsFix.fs
```diff
@@ -0,0 +1,43 @@
...
+
+            ModificationUtil.ReplaceChildRange(
```
auduchinok, line 40: We expect the subexpressions to be in a `sequenceExpr`, so we're breaking the tree with replacing the tuple with its contents.
https://github.com/JetBrains/fsharp-support/blob/net211/ReSharper.FSharp/test/data/parsing/Expr%20-%20List%20-%20Seq%2001.fs.gold#L12

Btw, is it safe to replace a parent node with its child without copying here?

ReSharper.FSharp/src/FSharp.Psi/src/Impl/Tree/ArrayOrListExpr.cs
auduchinok: @DedSec256 Thanks for updating the psi!

ReSharper.FSharp/src/FSharp.Psi.Features/src/QuickFixes/ConvertTupleToArrayOrListElementsFix.fs
DedSec256: It is planned to add support for other warnings in the future. But okay, this constructor can be removed now.

ReSharper.FSharp/src/FSharp.Psi.Features/src/QuickFixes/ConvertTupleToArrayOrListElementsFix.fs
DedSec256: good idea, thanks!

DedSec256: But we also have the non-standard case ``[(a, b, c)]`` - it is not enough just to replace the separators with ';' but also get rid of the explicit parentheses

ReSharper.FSharp/src/FSharp.Psi.Features/src/QuickFixes/ConvertTupleToArrayOrListElementsFix.fs
DedSec256: I manually verified that after applying the fix, the tree structure remains correct. You can also explain the correctness by the fact that the sequence elements and ';' separators are recalculated from children on each call. 
image

ReSharper.FSharp/src/FSharp.Psi.Features/src/QuickFixes/ConvertTupleToArrayOrListElementsFix.fs
auduchinok: Good point! I'm OK with either ignoring such cases or removing the parens.

ReSharper.FSharp/src/FSharp.Psi.Features/src/QuickFixes/ConvertTupleToArrayOrListElementsFix.fs
auduchinok: > I manually verified that after applying the fix, the tree structure remains correct. 

Please show dumps of the trees for `[1;2]` and for before/after the quick fix on `[1,2]`. The tree after the quick fix is expected to be the same as `[1;2]`.

DedSec256: Oh, yes, indeed, there is no ``sequenceExpr`` at all, sorry. I will fix it.

auduchinok (approved):

PR #221, opened 2021-01-16
Add 'To interpolated string' analyzer/quick fix

saul:
Analyzer to convert a printf-style format string to an F# 5 interpolated string.

interpolate-better


slang25:
Give it to me now!

ReSharper.FSharp/src/FSharp.Psi/src/IFSharpElementFactory.cs
```diff
@@ -52,5 +52,7 @@ public interface IFSharpElementFactory
     IInterfaceImplementation CreateInterfaceImplementation(ITypeReferenceName typeReferenceName, FSharpList<IMemberDeclaration> memberDeclarations, int indent);
 
     ITypeParameterOfTypeList CreateTypeParameterOfTypeList(FSharpList<string> names);
+
+    IInterpolatedStringExpr CreateInterpolatedString(string contents);
```
auduchinok, line 56: I think it'd be OK to reuse CreateExpr here.

ReSharper.FSharp/src/FSharp.Psi.Features/src/Intentions/ToInterpolatedString.fs
```diff
@@ -0,0 +1,155 @@
...
+
+        let literalExpr = prefixAppExpr.ArgumentExpression.IgnoreInnerParens().As<ILiteralExpr>()
+        if isNull literalExpr then false else
+
+        // todo: support FSharpTokenType.TRIPLE_QUOTED_STRING and FSharpTokenType.VERBATIM_STRING
+        let tokenType = literalExpr.Literal.GetTokenType()
+        if tokenType <> FSharpTokenType.STRING then false else
+
+        let sourceFile = prefixAppExpr.GetSourceFile()
+        match prefixAppExpr.CheckerService.ParseAndCheckFile(sourceFile, opName, true) with
```
auduchinok, line 44: Please use `IFSharpFile.GetParseAndCheckResults` instead.

ReSharper.FSharp/src/FSharp.Psi.Features/src/Intentions/ToInterpolatedString.fs
```diff
@@ -0,0 +1,155 @@
...
+                Description = "Convert an printf-style format string with an interpolated string")>]
+type ToInterpolatedStringAction(dataProvider: FSharpContextActionDataProvider) =
+    inherit FSharpContextActionBase(dataProvider)
+
+    let [<Literal>] opName = "ToInterpolatedStringAction"
+
+    override x.Text = "To interpolated string"
+
+    override x.IsAvailable _ =
+        let prefixAppExpr = dataProvider.GetSelectedElement<IPrefixAppExpr>()
```
auduchinok, line 30: I think it should only be available when caret is at `sprintf`-like function identifier or at the corresponding string literal. Making it available in more places (e.g. at the `sprintf` argument) will likely interfere with other actions in future. Take a look at the existing actions for examples.

ReSharper.FSharp/src/FSharp.Psi.Features/src/Intentions/ToInterpolatedString.fs
```diff
@@ -0,0 +1,155 @@
...
+        if isNull prefixAppExpr then false else
+
+        // todo: fix NRE in tests
+        //if not (FSharpLanguageLevel.isFSharp50Supported prefixAppExpr) then false else
+
+        let literalExpr = prefixAppExpr.ArgumentExpression.IgnoreInnerParens().As<ILiteralExpr>()
+        if isNull literalExpr then false else
+
+        // todo: support FSharpTokenType.TRIPLE_QUOTED_STRING and FSharpTokenType.VERBATIM_STRING
+        let tokenType = literalExpr.Literal.GetTokenType()
```
auduchinok, line 40: You should always assume that generated child node properties like `Literal` may return `null` due to error in file. There's a `getTokenType` function that does the same while is safe to use here.

ReSharper.FSharp/src/FSharp.Psi.Features/src/Intentions/ToInterpolatedString.fs
```diff
@@ -0,0 +1,155 @@
...
+            )
+            |> Seq.filter (fun (range, _) -> argRange.Contains(&range))
+            |> Array.ofSeq
+
+        // Find the outermost IPrefixAppExpr and all applied exprs (excluding the format string itself)
+        let outerPrefixAppExpr, appliedExprs =
+            let rec loop acc (expr: IPrefixAppExpr) =
+                match PrefixAppExprNavigator.GetByFunctionExpression (expr.IgnoreParentParens()) with
+                | null -> expr.IgnoreParentParens(), acc
+                | parent -> loop (parent.ArgumentExpression::acc) parent
```
auduchinok, line 90: Please add spaces around `::`.

auduchinok (commented):
Just a few comments, I'll look closer at the tree modification later and will do another review.

ReSharper.FSharp/src/FSharp.Psi.Features/src/Intentions/ToInterpolatedString.fs
saul: I need to add the availability tests, but at the moment it's only available on the `sprintf` or the format string itself - not anywhere else.

saul: Just pushed some availability tests

ReSharper.FSharp/src/FSharp.Psi.Features/src/QuickFixes/ReplaceWithInterpolatedStringFix.fs
```diff
@@ -0,0 +1,96 @@
...
+                    .Insert(index + 1, exprText)
+                    .Insert(index + 1 + exprText.Length, '}')
+                |> ignore
+
+        let factory = formatStringExpr.CreateElementFactory()
+
+        interpolatedSb.Insert(0, '$') |> ignore
+        let interpolatedStringExpr = factory.CreateExpr(interpolatedSb.ToString())
+
+        // todo: get this working with isPredefinedFunctionRef
```
saul, line 89: @auduchinok any ideas here would be appreciated. This used to work when it was an intention, but now doesn't work as an analyzer 😕

auduchinok: If that's with referenced `FSharp.Core/5.0`, then this package won't currently restore in tests, so R# wouldn't be able to find symbols in it either.

drhumlen:
This would be really nice to have in 2021.1 – any hope of getting this merged @saul @auduchinok ?

saul:
Resolved conflicts, test failures and todo. @auduchinok this is ready now.

auduchinok:
@saul Thanks! Will review shortly!

ReSharper.FSharp/src/FSharp.Psi.Features/src/Daemon/Analyzers/InterpolatedStringCandidateAnalyzer.fs
```diff
@@ -0,0 +1,95 @@
...
+            let isDisallowed =
+                if tokenType = FSharpTokenType.TRIPLE_QUOTED_STRING then
+                    fun (literalExpr: ILiteralExpr) ->
+                        getTokenType literalExpr.Literal = FSharpTokenType.TRIPLE_QUOTED_STRING
+                else
+                    isDisallowedStringLiteral
+
+            appliedExprs
+            |> List.exists (fun expr ->
+                expr.ThisAndDescendants<ILiteralExpr>().ToEnumerable() |> Seq.exists isDisallowed)
```
auduchinok, line 83: Could be cool to convert the resulting string into triple quote one (assuming we can guarantee it's safe in particular cases) in future, so inner regular strings were allowed. Also if some arguments are simple string literals, they could be embedded into resulting string directly.

ReSharper.FSharp/test/data/features/daemon/interpolatedStringCandidate/Not available 07 - Other string.fs.gold
```diff
@@ -0,0 +1,7 @@
+﻿module Module
+
+let foo = id "foo"
+let s = sprintf |"%d %s"|(0) 5 foo
+
+---------------------------------------------------------
+(0): ReSharper Suggestion: Format string can be replaced with an interpolated string
```
auduchinok, line 7: I wonder why it's called Not available? 🙂

ReSharper.FSharp/test/data/features/daemon/interpolatedStringCandidate/Not available 13 - Descendant string in applied arg.fs
```diff
@@ -0,0 +1,4 @@
+module Module
+
+let d = System.DateTime.Now
+let s = sprintf "%d %s" 5 (d.ToString("d"))
```
auduchinok, line 4: Very nice check, thank you!

ReSharper.FSharp/test/data/features/quickFixes/replaceWithInterpolatedString/String 01 - Single specifier.fs.gold
```diff
@@ -0,0 +1,3 @@
+﻿module Module
+
+let s = ${caret}"%d{10}"
```
auduchinok, line 3: Can we assume it's safe to remove specifiers when an inserted expression is a literal expression?

auduchinok (approved):
@saul This is very nice! Very good test coverage and everything just looks robust and ready to go. Thank you!

ReSharper.FSharp/test/data/features/daemon/interpolatedStringCandidate/Not available 07 - Other string.fs.gold
saul: Think it was for the `let foo = ...` line above - although in hindsight it's a bit of a pointless test.

vasily-kirichenko:
I think redundant specifiers should be deleted while refactoring. For example, `"%s %d" s d` where `s` is a `string` and `d` is an `int` should be transformed to `$"{s} {d}`, not to `$%s{s} %d{d}`. Currently only `%O` is removed, which is nice.

NinoFloris:
I have thought about this as well however maybe you really only want to do that if string/int/etc is rigid and not inferred, otherwise you lose the upside of the typed sprintf/specifiers.

saul:
Yes exactly. I don't think there's a way to know whether removing a specifier is safe/won't affect type inference.

slang25:
I've been using this feature and loving it! I like it keeps the type prefixes, and I can remove them only when it makes sense to do so.

PR #224, opened 2021-01-25
Downgrade C# LangVersion in the project to 8.0

DedSec256:


auduchinok (changes requested):
Could you update the second C# project too?

auduchinok (approved):
Thanks!

PR #226, opened 2021-01-31
Enable formatter in more quick fixes

DedSec256:


DedSec256:
It seems that this case with the enabled formatter 
image

will lead to an error 

image

auduchinok:
@DedSec256 We should not enable it then yet.

PR #230, opened 2021-02-17
Use a constant defined in R#

kirillgla:
This will fix a trailing `_Rider` from tooltips that use the separator


auduchinok (approved):
Thanks!

PR #231, opened 2021-02-20
Add Rider self debug configurations for the backend solution

ForNeVeR:
@auduchinok, could you check if it works on Mac, please?

Obviously, we could try to do better and create a single configuration for both Unix and Windows (in form of the combined sh/cmd script), but for start, let's check if it works at all :)

(I've tested that on Windows)

### DIFF
--- a/ReSharper.FSharp/.idea/.idea.ReSharper.FSharp/.idea/runConfigurations/Rider_Debug__Unix_.xml
+++ b/ReSharper.FSharp/.idea/.idea.ReSharper.FSharp/.idea/runConfigurations/Rider_Debug__Unix_.xml
@@ -1,0 +1,17 @@
+<component name="ProjectRunConfigurationManager">
+  <configuration default="false" name="Rider Debug (Unix)" type="RunExe" factoryName=".NET Executable">
+    <option name="EXE_PATH" value="$PROJECT_DIR$/../rider-fsharp/gradlew" />
+    <option name="PROGRAM_PARAMETERS" value="runIde" />
+    <option name="WORKING_DIRECTORY" value="$PROJECT_DIR$/../rider-fsharp" />
+    <option name="PASS_PARENT_ENVS" value="1" />
+    <envs>
+      <env name="RESHARPER_HOST_DEBUG_SELF" value="1" />
+    </envs>
+    <option name="USE_EXTERNAL_CONSOLE" value="0" />
+    <option name="USE_MONO" value="0" />
+    <option name="RUNTIME_ARGUMENTS" value="" />
+    <method v="2">
+      <option name="Build Solution" enabled="true" />
+    </method>
+  </configuration>
+</component>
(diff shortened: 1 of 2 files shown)


auduchinok:
@ForNeVeR This is what I get on Run:
```
/Library/Frameworks/Mono.framework/Versions/6.8.0/bin/mono-sgen64 /Users/eugene/Developer/resharper-fsharp/rider-fsharp/gradlew runIde
Cannot open assembly '/Users/eugene/Developer/resharper-fsharp/rider-fsharp/gradlew': File does not contain a valid CIL image.

Process finished with exit code 2.
```

This is what happens why I try to Debug:
Screenshot 2021-02-20 at 15 36 27

ReSharper.FSharp/.idea/.idea.ReSharper.FSharp/.idea/runConfigurations/Rider_Debug__Unix_.xml
```diff
@@ -0,0 +1,17 @@
+<component name="ProjectRunConfigurationManager">
+  <configuration default="false" name="Rider Debug (Unix)" type="RunExe" factoryName=".NET Executable">
```
auduchinok, line 2: With the following change I'm able to launch the IDE but something is very off with it: daemon doesn't work, there're no exceptions, everything just doesn't seem to be working properly.
```suggestion
  <configuration default="false" name="Rider Debug (Unix)" type="RunNativeExe" factoryName="Native Executable">
```
<configuration default="false" name="Rider Debug (Unix)" type="RunNativeExe" factoryName="Native Executable">

auduchinok: OK, with the latest Rider dev build debugging does work, but it still requires the change above. @ForNeVeR @citizenmatt Is it expected?

citizenmatt: Yes. The .NET Executable run configuration on Mac will try to launch the `gradlew` shell script as a Mono executable, and so you get the `File does not contain a valid CIL image` error. You need to use the `RunNativeExe` config to launch a shell script, and it's what we're doing in the product, too.

auduchinok (approved):

ReSharper.FSharp/.idea/.idea.ReSharper.FSharp/.idea/runConfigurations/Rider_Debug__Unix_.xml
ForNeVeR: Thanks, guys. I wasn't aware of that issue with Mono (probably this is something that will be improved in 2021.2, though).

PR #235, opened 2021-02-23
Escape newlines in tooltips

kirillgla:
We need to do this on the F# side because if we do this on the Rider side we'd have to escape all of the tooltips, which would cause us to insert `<br>` tags inside CSS, which totally breaks a lot of tooltips


auduchinok (approved):
Thanks!

PR #238, opened 2021-03-01
Add ReplaceWithPredefinedOperatorFix

DedSec256:


ReSharper.FSharp/src/FSharp.Psi.Features/src/Daemon/Stages/FcsErrorsStageProcessBase.fs
```diff
@@ -171,6 +172,9 @@ type FcsErrorsStageProcessBase(fsFile, daemonProcess) =
 
             UnresolvedHighlighting(error.Message, range) :> _
 
+        | ErrorFromAddingConstraint ->
+            createHighlightingFromNodeWithMessage ErrorFromAddingConstraintError range error
```
auduchinok, line 176: Why does it need two `Error` in its name?

ReSharper.FSharp/src/FSharp.Psi.Features/src/QuickFixes/ReplaceWithInequalityOperatorFix.fs
```diff
@@ -0,0 +1,27 @@
...
+        | _ -> false
+
+    override x.Text = "Replace with '<>'"
+
+    override x.ExecutePsiTransaction _ =
+        use writeCookie = WriteLockCookie.Create(expr.IsPhysical())
+        use disableFormatter = new DisableCodeFormatter()
+        let factory = expr.CreateElementFactory()
+
+        replace expr (factory.CreateReferenceExpr("op_Inequality").SetName("<>"))
```
auduchinok, line 27: This line looks a bit questionable, since you don't need the passed and created reference name right after creating it.

auduchinok (commented):
Looks almost fine. Would it be easy to add `==` as well?

ReSharper.FSharp/src/FSharp.Psi.Features/FSharp.Psi.Features.fsproj
```diff
@@ -200,7 +200,7 @@
     <Compile Include="src\QuickFixes\ReplaceWithAbbreviatedTypeFix.fs" />
     <Compile Include="src\QuickFixes\UseWildSelfIdFix.fs" />
     <Compile Include="src\QuickFixes\SpecifyParameterTypeFix.fs" />
-    <Compile Include="src\QuickFixes\ReplaceWithInequalityOperatorFix.fs" />
+    <Compile Include="src\QuickFixes\ReplaceWithFSharpOperatorFix.fs" />
```
auduchinok, line 203: I wonder where `FSharp` comes from? `==` and `!=` are totally valid F# operators. Just are not predefined/built-in.

ReSharper.FSharp/src/FSharp.Psi.Features/src/Daemon/Highlightings/FcsErrors.xml
```diff
@@ -328,15 +328,15 @@
     <QuickFix>SpecifyParameterTypeFix</QuickFix>
   </Error>
 
-  <Error staticGroup="FSharpErrors" name="ErrorFromAddingConstraint">
+  <Error staticGroup="FSharpErrors" name="ConstraintAdding">
```
auduchinok, line 331:
```suggestion
  <Error staticGroup="FSharpErrors" name="AddingConstraint">
```

ReSharper.FSharp/src/FSharp.Psi.Features/src/QuickFixes/ReplaceWithFSharpOperatorFix.fs
```diff
@@ -0,0 +1,36 @@
...
+open JetBrains.ReSharper.Plugins.FSharp.Psi.Features.Daemon.QuickFixes
+open JetBrains.ReSharper.Plugins.FSharp.Psi.Tree
+open JetBrains.ReSharper.Psi.ExtensionsAPI
+open JetBrains.ReSharper.Resources.Shell
+
+type ReplaceWithFSharpOperatorFix(error: ConstraintAddingError) =
+    inherit FSharpQuickFixBase()
+
+    let ref = error.Expr.As<IReferenceExpr>()
+    let fsharpOperator =
```
auduchinok, line 14: Please add an empty line before the multiline binding.

ReSharper.FSharp/src/FSharp.Psi.Features/src/QuickFixes/ReplaceWithFSharpOperatorFix.fs
```diff
@@ -0,0 +1,36 @@
...
+        match ref with
+        | null -> ""
```
auduchinok, line 16: Why not use the missing name literal?

ReSharper.FSharp/src/FSharp.Psi.Features/src/QuickFixes/ReplaceWithFSharpOperatorFix.fs
```diff
@@ -0,0 +1,36 @@
...
+        | ref ->
+
+        match ref.ShortName with
+        | "==" -> "="
+        | "!=" -> "<>"
+        | _ -> ""
+
+    override x.IsAvailable _ = isValid ref && fsharpOperator <> ""
```
auduchinok, line 24: Please check that the reference is in a binary app expression.
Also, there seems to be no test scenario like this: Screenshot 2021-03-01 at 18 10 49

DedSec256: I think the check for IBinaryAppExpr is unnecessary because we can apply the fix to more cases such as 
`fun x -> (==) 1`

auduchinok: Then we should take into account possible qualifier expressions and/or type argument lists, so setting the reference name only is likely not enought.

auduchinok: Also, if we're to support prefix app expression cases, please add a test case. 🙂

DedSec256:
At the moment the scope of the fix is ​​IBinaryAppExpr

ReSharper.FSharp/src/FSharp.Psi.Features/src/QuickFixes/ReplaceWithPredefinedOperatorFix.fs
```diff
@@ -0,0 +1,37 @@
...
+open JetBrains.ReSharper.Plugins.FSharp.Psi.Tree
+open JetBrains.ReSharper.Psi.ExtensionsAPI
+open JetBrains.ReSharper.Resources.Shell
+
+type ReplaceWithPredefinedOperatorFix(error: AddingConstraintError) =
+    inherit FSharpQuickFixBase()
+
+    let ref = error.Expr.As<IReferenceExpr>()
+
+    let fsharpOperator =
```
auduchinok, line 15: What does `fsharp` stand for here?

ReSharper.FSharp/src/FSharp.Psi.Features/src/QuickFixes/ReplaceWithPredefinedOperatorFix.fs
```diff
@@ -0,0 +1,37 @@
...
+        match ref with
+        | null -> SharedImplUtil.MISSING_DECLARATION_NAME
+        | ref when ref.IsQualified -> SharedImplUtil.MISSING_DECLARATION_NAME
```
auduchinok, line 18: So if there's a name and a qualifier then you'll say that name is missing?

ReSharper.FSharp/src/FSharp.Psi.Features/src/QuickFixes/ReplaceWithPredefinedOperatorFix.fs
DedSec256: renaming lost before submitting the commit :(

ReSharper.FSharp/src/FSharp.Psi.Features/src/QuickFixes/ReplaceWithPredefinedOperatorFix.fs
DedSec256: I think this line is not needed at all, since qualified access to the operator presupposes a prefix notation

auduchinok (approved):
LGTM, thanks! 🙂

PR #245, opened 2021-03-18
TypeProviders: share ProvidedTypeContext between generated types

DedSec256:
Fix for https://youtrack.jetbrains.com/issue/RIDER-59965


auduchinok (commented):
@DedSec256 Could you add a test, please?

DedSec256:
Changed SimpleGenerativeProvider.fs сovers the case where we are trying to access the nested type via the root type abbreviation (and when a separate abbreviation has not yet been declared for the nested type). 

Before the fix, we would get the error that a separate abbreviation for the nested type is required, which is not correct.

auduchinok (approved):

PR #252, opened 2021-04-05
Move Fantomas out-of-process [<WIP>]

DedSec256:


nojaf:
Hey Alex, this looks interesting! Could you elaborate a bit more on what the main idea is here?
I assume the end goal is to be more flexible in updating Fantomas.
Would the user be able to bring their own version of Fantomas?

I'm asking because I was messing around with [a similar idea](https://github.com/fsprojects/fantomas/compare/master...nojaf:ras-al-ghul#diff-c7793bb4882787f0586020d135b7821b4907482f7c178b8bb5dee13b3e962708R56). 
I had in mind to be able to start a Fantomas process as a language server and keep it running.
Editor tooling could then connect to this and format on demand.
The idea is that the version pinned in the [tool manifest](https://docs.microsoft.com/en-us/dotnet/core/tools/local-tools-how-to-use#create-a-manifest-file) of the user could be used by the editor.

Perhaps some synergy is possible between this PR and my experiment.

DedSec256:
Hi @nojaf!
This pull request contains the preparatory part required to run Fantomas and its dependencies in a separate process (currently the bundled by default version).

Yes, in the future it is planned to add support for pinned custom Fantomas version in a separate PR (:
Thanks for the link to your experiment, I'll take a look at this.

ReSharper.FSharp/src/FSharp.ExternalFormatter.Protocol/src/ExternalFormatterProcess.cs
```diff
@@ -0,0 +1,71 @@
...
+
+    protected override ProcessStartInfo GetProcessStartInfo(int port)
+    {
+      var launchPath = GetType().Assembly.GetPath().Directory.Combine(ProtocolConstants.PROCESS_FILENAME);
+      Assertion.Assert(launchPath.ExistsFile, $"can't find '{ProtocolConstants.PROCESS_FILENAME}'");
+
+      return new ProcessStartInfo
+      {
+        Arguments =
+          $"{port} \"{ProtocolConstants.LogFolder.Combine($"{DateTime.UtcNow:yyyy_MM_dd_HH_mm_ss_ffff}.log")}\"",
```
auduchinok, line 39: Should it mention something about Fantomas host or F#? Looks like a common class at a glance.

ReSharper.FSharp/src/FSharp.Psi.Features/src/Formatter/FSharpReformatCode.fs
```diff
@@ -39,83 +36,44 @@ type FSharpReformatCode() =
...
                     try
                         let range = ofDocumentRange rangeMarker.DocumentRange
 
                         let formatted =
-                            CodeFormatter
-                                .FormatSelectionAsync(filePath, range, source, formatConfig, parsingOptions, checker)
-                                .RunAsTask()
+                            codeFormatterProvider
+                                .FormatSelection(filePath, range, text, settings, parsingOptions)
                                 .Replace("\r\n", newLineText)
```
auduchinok, line 68: I wonder if it would be better to move this result processing the the other process too?

rider-fsharp/protocol/src/kotlin/model/RdFSharpExternalFormatterModel.kt
```diff
@@ -0,0 +1,72 @@
...
+import com.jetbrains.rider.model.nova.ide.SolutionModel
+import com.jetbrains.rd.generator.nova.*
+import com.jetbrains.rd.generator.nova.PredefinedType.*
+import com.jetbrains.rd.generator.nova.csharp.CSharp50Generator
+import java.io.File
+
+@Suppress("unused")
+object RdFSharpExternalFormatterModel : Root() {
+
+    private val rdParsingOptions = structdef {
```
auduchinok, line 12: It seems to miss a lot of info from the parsing options. Something like the following will likely be formatted wrong:
```fsharp
#if DEBUG
let x = 123
#endif
```

rider-fsharp/protocol/src/kotlin/model/RdFSharpExternalFormatterModel.kt
```diff
@@ -0,0 +1,72 @@
...
+
+    private val rdRange = structdef {
+        field("fileName", string)
+        field("startLine", int)
+        field("startCol", int)
+        field("endLine", int)
+        field("endCol", int)
+    }
+
+    private val rdFormatConfig = structdef {
```
auduchinok, line 25: Name suggestion: `rdFantomasFormatConfig`.

rider-fsharp/protocol/src/kotlin/model/RdFSharpExternalFormatterModel.kt
auduchinok: Name suggestion: `rdFcsParsingOptions`.

rider-fsharp/protocol/src/kotlin/model/RdFSharpExternalFormatterModel.kt
```diff
@@ -0,0 +1,72 @@
...
+
+@Suppress("unused")
+object RdFSharpExternalFormatterModel : Root() {
+
+    private val rdParsingOptions = structdef {
+        field("sourceFiles", array(string))
+        field("lightSyntax", bool.nullable)
+    }
+
+    private val rdRange = structdef {
```
auduchinok, line 17: Name suggestion: `rdFcsRange`.

auduchinok:
@DedSec256 Could you also please add tests or either/both backend and frontend?

The start looks good!

ReSharper.FSharp/src/FSharp.ExternalFormatter.Protocol/src/ExternalFormatterProcess.cs
DedSec256: pid of the parent process is passed in [environment variables](https://github.com/JetBrains/resharper-fsharp/pull/252/files/92d10c62dd2f875243d5294878ea0643248e6c77#diff-2e386d439455966e46f2935fbdff0b0df204fe948b9fd71427d3916b71e6a446R54)

ReSharper.FSharp/src/FSharp.ExternalFormatter.Host/src/ExternalFormatterEndPoint.cs
```diff
@@ -0,0 +1,53 @@
...
+
+    protected override void InitLogger(Lifetime lifetime, string path)
+    {
+      LogManager.Instance.SetConfig(new XmlLogConfigModel());
+      var logPath = FileSystemPath.TryParse(path);
+      if (logPath.IsNullOrEmpty()) return;
+
+      var logEventListener = new FileLogEventListener(logPath);
+      LogManager.Instance.AddOmnipresentLogger(lifetime, logEventListener, LoggingLevel.TRACE);
+    }
```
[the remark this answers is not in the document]
DedSec256: This is an abstract method, everyone implements it, but I think in the plugin we can create separate Utils for the generalized logger configuration (for the formatter and type providers)

ReSharper.FSharp/src/FSharp.ExternalFormatter.Protocol/src/ExternalFormatterProcess.cs
auduchinok: ?

ReSharper.FSharp/src/FSharp.ExternalFormatter.Host/src/ExternalFormatterEndPoint.cs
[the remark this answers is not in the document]
auduchinok: Do these implementations have different and meaningful logic? Would it benefit other implementations if we add such an util to the platform?

ReSharper.FSharp/src/FSharp.ExternalFormatter.Host/src/CodeFormatterProvider.cs
```diff
@@ -0,0 +1,61 @@
...
+using FSharp.Compiler.CodeAnalysis;
+using FSharp.Compiler.Diagnostics;
+using FSharp.Compiler.Text;
+using JetBrains.Rider.FSharp.ExternalFormatter.Client;
+using Microsoft.FSharp.Collections;
+using Microsoft.FSharp.Control;
+
+namespace FSharp.ExternalFormatter.Host
+{
+  internal interface ICodeFormatterProvider
```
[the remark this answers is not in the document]
DedSec256: initially, it seemed to me that custom versions of Fantomas require a separate implementation via reflection, but it seems we won't need this, yes (:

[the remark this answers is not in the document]
auduchinok: Anyway, even if it used reflection, would there be anything other than Fantomas? None of R# formatters would work here, so the name would still be misleading even with reflection.

ReSharper.FSharp/src/FSharp.ExternalFormatter.Protocol/src/ExternalFormatterProcess.cs
auduchinok: I've tried to clarify my comment.

ReSharper.FSharp/src/FSharp.ExternalFormatter.Host/src/ExternalFormatterEndPoint.cs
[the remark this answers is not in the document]
DedSec256: yes, I think we can make a generic implementation in the platform

ReSharper.FSharp/src/FSharp.ExternalFormatter.Protocol/src/ExternalFormatterProcess.cs
DedSec256: nobody in the platform did this because the namespace already contains the necessary information. But we can (:

ReSharper.FSharp/src/FSharp.ExternalFormatter.Host/FSharp.ExternalFormatter.Host.csproj
```diff
@@ -0,0 +1,26 @@
...
+    </ItemGroup>
+
+    <ItemGroup>
+      <PackageReference Include="JetBrains.Fantomas" Version="$(FantomasVersion)" />
+    </ItemGroup>
+
+    <ItemGroup>
+      <None Update="JetBrains.ReSharper.Plugins.FSharp.ExternalFormatter.Host.runtimeconfig.json">
+        <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+      </None>
```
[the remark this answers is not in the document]
DedSec256: We are currently [using](https://github.com/JetBrains/resharper-fsharp/blob/6a31403c7767349bc4165b50324c14ef7e3e2080/ReSharper.FSharp/src/FSharp.ExternalFormatter.Protocol/src/ExternalFormatterProcessFactory.cs#L33) the internal rider runtime, 
but we still need to carry the runtimeconfig with us.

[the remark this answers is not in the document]
auduchinok: > but we still need to carry the runtimeconfig with us.

Could you please tell why?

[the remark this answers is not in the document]
DedSec256: The platform doesn't yet have the necessary launchers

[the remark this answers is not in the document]
auduchinok: OK, thanks! Is there an issue to track/vote?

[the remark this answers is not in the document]
DedSec256: It looks like this is it
https://youtrack.jetbrains.com/issue/RIDER-59978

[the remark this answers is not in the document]
auduchinok: Cool, thanks! 🙂

ReSharper.FSharp/src/FSharp.ExternalFormatter.Host/src/ExternalFormatterAssemblyResolver.cs
```diff
@@ -0,0 +1,43 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Reflection;
+
+namespace FSharp.ExternalFormatter.Host
+{
+  public static class ExternalFormatterAssemblyResolver
+  {
+    private const string AdditionalProbingPathsEnvVar = "RIDER_PLUGIN_ADDITIONAL_PROBING_PATHS";
```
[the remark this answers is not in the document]
DedSec256: This constant defines the environment variable passed by the Rider to all child processes. And yes, it is not in the platform :D

ReSharper.FSharp/src/FSharp.ExternalFormatter.Host/FSharp.ExternalFormatter.Host.csproj
```diff
@@ -0,0 +1,27 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <TargetFramework>net461</TargetFramework>
+        <AssemblyName>JetBrains.ReSharper.Plugins.FSharp.ExternalFormatter.Host</AssemblyName>
+        <OutputType>Exe</OutputType>
+        <SignAssembly>false</SignAssembly>
```
auduchinok, line 7: Why this line is needed?

ReSharper.FSharp/src/FSharp.ExternalFormatter.Host/FSharp.ExternalFormatter.Host.csproj
```diff
@@ -0,0 +1,27 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <TargetFramework>net461</TargetFramework>
+        <AssemblyName>JetBrains.ReSharper.Plugins.FSharp.ExternalFormatter.Host</AssemblyName>
```
auduchinok, line 5: Please fix the `ExternalFormatter` vs `Fantomas` naming.

rider-fsharp/protocol/src/kotlin/model/RdFSharpExternalFormatterModel.kt
```diff
@@ -0,0 +1,73 @@
...
+        field("newLineText", string)
+    }
+
+    init {
+        call("formatDocument", structdef("rdFormatDocumentArgs") extends rdFormatArgs {}, string)
+        call("formatSelection", structdef("rdFormatSelectionArgs") extends rdFormatArgs {
+            field("range", rdFcsRange)
+        }, string)
+
+        signal("Exit", void)
```
auduchinok, line 71: Please fix the casing.

ReSharper.FSharp/src/FSharp.ExternalFormatter.Host/FSharp.ExternalFormatter.Host.csproj
DedSec256: in this pull request, we may not need it, but we will need it to load the custom Fantomas version.

auduchinok: In what cases this line could be needed? Isn't it the default value?

DedSec256: yes you seem to be right, it is false by default

ReSharper.FSharp/src/FSharp.ExternalFormatter.Host/src/ExternalFormatterAssemblyResolver.cs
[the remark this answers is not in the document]
auduchinok: BTW, what does "rider plugin" mean in the literal?

ReSharper.FSharp/src/FSharp.Fantomas.Host/FSharp.Fantomas.Host.csproj
```diff
@@ -8,6 +8,14 @@
         <RootNamespace>JetBrains.ReSharper.Plugins.FSharp.Fantomas.Host</RootNamespace>
     </PropertyGroup>
 
+    <PropertyGroup Condition=" '$(Configuration)' == 'Debug' ">
+      <PlatformTarget>AnyCPU</PlatformTarget>
```
auduchinok, line 12: Isn't it the default platform target? Assuming yes, is there any other place it gets overridden?

auduchinok: I've just checked,  and `dotnet new console` creates `AnyCPU` for me:
Screenshot 2021-05-11 at 15 52 11

DedSec256: Fantomas.Host.exe targets .NET 4.6.1, there is x86 by default

auduchinok (changes requested):
@DedSec256 Could you finish the remaining bits, please?

ReSharper.FSharp/src/FSharp.ExternalFormatter.Host/src/CodeFormatterProvider.cs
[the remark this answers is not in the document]
DedSec256: fixed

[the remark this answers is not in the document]
auduchinok: From a glance, there're some remaining bits:

https://github.com/JetBrains/resharper-fsharp/pull/252/files#diff-275de39244e639c3bb36c6756aae538f325b8bc9ba079dfc204741f6b6fde1ceR108

https://github.com/JetBrains/resharper-fsharp/pull/252/files#diff-ffd401a725f47038970593e5f58d09e68d56e5172a6cc9b99148477e2b484b1eR13

auduchinok (approved):

PR #253, opened 2021-04-05
Add ReplaceIfWithConditionOperand analyzer

DedSec256:
JetBrains


saul:
What if there are comments in the branches? e.g.

```f#
if true then
    // Do something
    true
else
    // Do something else
    false
```

DedSec256:
@saul, for now, the analyzer and quickfix will suggest replacing such an expression. Do you think it's better not to suggest a fix for such cases?

ReSharper.FSharp/src/FSharp.Psi.Features/src/Daemon/Highlightings/FSharpErrorUtil.fs
```diff
@@ -116,6 +116,10 @@ let getLambdaCanBeReplacedWarningText (replaceCandidate: IFSharpExpression) =
         sprintf "Lambda can be replaced with '%s'" x.QualifiedName
     | _ -> "Lambda can be simplified"
 
+let getIfCanBeReplacedByConditionOperandWarningText needNegation =
+    if needNegation then "'if' expression can be replaced by condition operand negation"
+    else "'if' expression can be replaced by condition operand"
```
auduchinok, line 121: `with` seems to suit better than `by` to me, although I'm not a native speaker.
https://english.stackexchange.com/questions/151309/replace-with-versus-replace-by

ReSharper.FSharp/src/FSharp.Psi.Features/src/Daemon/Analyzers/IfExpressionAnalyzer.fs
```diff
@@ -0,0 +1,36 @@
...
+[<ElementProblemAnalyzer([| typeof<IIfThenElseExpr> |],
+                         HighlightingTypes = [| typeof<IfCanBeReplacedByConditionOperandWarning> |])>]
+type IfExpressionAnalyzer() =
+    inherit ElementProblemAnalyzer<IIfThenElseExpr>()
+
+    override this.Run(expr, _, consumer) =
+        if not (skipMatchingNodesAfter isWhitespace expr.ThenKeyword == expr.ThenExpr) ||
+           not (skipMatchingNodesAfter isWhitespace expr.ThenExpr == expr.ElseKeyword) ||
+           not (skipMatchingNodesAfter isWhitespace expr.ElseKeyword == expr.ElseExpr) then () else
+
```
auduchinok, line 20: Do we need to traverse all the expressions surroundings before even checking if we're interested in this `if` expression?

auduchinok: This code will fail with NREs as soon as any of the accessed children is missing, breaking the rest of the analysis too. There should be a test with missing `else` branch, and every child accessor should have a `null` check.

ReSharper.FSharp/src/FSharp.Psi.Features/src/Daemon/Analyzers/IfExpressionAnalyzer.fs
```diff
@@ -0,0 +1,36 @@
...
+        let thenExpr = expr.ThenExpr.IgnoreInnerParens()
+        let elseExpr = expr.ElseExpr.IgnoreInnerParens()
+
+        match thenExpr, elseExpr with
+        | :? ILiteralExpr as thenExpr, (:? ILiteralExpr as elseExpr) ->
+            let thenToken = getTokenType thenExpr.Literal
+            let elseToken = getTokenType elseExpr.Literal
+
+            if thenToken = FSharpTokenType.TRUE &&
```
auduchinok, line 29: Use `==` for checking reference equality with singletons.

auduchinok (changes requested):
This PR is very dangerous and will easily break.

ReSharper.FSharp/src/FSharp.Psi.Features/src/Daemon/Highlightings/FSharpErrorUtil.fs
DedSec256: I did it by analogy with C#

image

auduchinok: I think this one would be more similar to our case:
Screenshot 2021-05-26 at 14 37 43

Though, it's sad it's not very consistent in C# support.

ReSharper.FSharp/src/FSharp.Psi.Features/src/Daemon/Analyzers/IfExpressionAnalyzer.fs
```diff
@@ -0,0 +1,39 @@
...
+type IfExpressionAnalyzer() =
+    inherit ElementProblemAnalyzer<IIfThenElseExpr>()
+
+    override this.Run(expr, _, consumer) =
+        let thenExpr = expr.ThenExpr
+        let elseExpr = expr.ElseExpr
+
+        match thenExpr.IgnoreInnerParens(), elseExpr.IgnoreInnerParens() with
+        | :? ILiteralExpr as thenLiteral, (:? ILiteralExpr as elseLiteral) ->
+            let thenToken = getTokenType thenLiteral.Literal
```
auduchinok, line 22: `token` name for a token type is misleading. Especially, since we work with `then` token just a few lines below, and this then-expression-token-type is actually not the then-token.

ReSharper.FSharp/src/FSharp.Psi.Features/src/Daemon/Highlightings/FSharpErrorUtil.fs
```diff
@@ -116,6 +116,10 @@ let getLambdaCanBeReplacedWarningText (replaceCandidate: IFSharpExpression) =
         sprintf "Lambda can be replaced with '%s'" x.QualifiedName
     | _ -> "Lambda can be simplified"
 
+let getIfCanBeReplacedWithConditionOperandWarningText needNegation =
+    if needNegation then "'if' expression can be replaced with condition operand negation"
+    else "'if' expression can be replaced with condition operand"
+
```
auduchinok, line 122: Do we really need to add another function to this util? It should not be used for such simple cases, or it'll become too messy otherwise. Do we really need two different texts here?

PR #267, opened 2021-05-25
Type Providers: fix race in ProvidedEntitiesCache

DedSec256:


ReSharper.FSharp/src/FSharp.TypeProviders.Protocol/src/Cache/IProvidedEntitiesCache.cs
```diff
@@ -1,67 +1,88 @@
...
+  public abstract class ProvidedEntitiesCacheBase<T, TKey, TParam> : IProvidedEntitiesCache<T, TKey, TParam>
+    where T : class
   {
     protected readonly TypeProvidersContext TypeProvidersContext;
-    protected readonly IDictionary<int, T> Entities;
-    protected readonly IDictionary<int, List<int>> EntitiesPerProvider;
+    protected readonly ConcurrentDictionary<TKey, T> Entities;
+    protected readonly IDictionary<int, List<TKey>> EntitiesPerProvider;
+    private readonly object myEntitiesPerProviderLockObj = new object();
+    private SpinWaitLock myLock;
```
ForNeVeR, line 32: _(Minor)_ Let's name it `myEntitiesLock`?

ReSharper.FSharp/src/FSharp.TypeProviders.Protocol/src/Cache/IProvidedEntitiesCache.cs
```diff
@@ -1,67 +1,88 @@
...
-        .GroupBy(t => t.key == ProvidedConst.DefaultId || Entities.ContainsKey(t.key));
+        .GroupBy(t => !KeyHasValue(t.key) || Entities.ContainsKey(t.key));
 
       foreach (var group in groups)
       {
         if (group.Key)
           foreach (var (key, i) in group)
-            entities[i] = GetOrCreate(key, typeProviderId, context);
+          {
+            if (!KeyHasValue(key)) return null;
```
ForNeVeR, line 75: This is very strange. Should this really return `null` as `T[]` from `GetOrCreateBatch`? Thought it should just do `continue`?

ForNeVeR (changes requested):

ReSharper.FSharp/src/FSharp.TypeProviders.Protocol/src/Cache/IProvidedEntitiesCache.cs
DedSec256: you are right, there is a logical typo here

ForNeVeR (approved):

auduchinok (approved):

PR #321, opened 2021-11-04
Tree builder: group XmlDocs

DedSec256:
- [x] Support local bindings


ReSharper.FSharp/src/FSharp.Psi/src/FSharp.psi
```diff
@@ -1670,6 +1670,9 @@ anonRecordField:
   COLON<FSHARP_COLON, Colon>
   typeUsage<TYPE, TypeUsage>;
 
+xmlDocBlock:
+  LPAREN<LPAREN, LParen>;
```
auduchinok, line 1674: Why?

ReSharper.FSharp/src/FSharp.Psi.Features/src/Parsing/FSharpImplTreeBuilder.fs
```diff
@@ -98,6 +95,13 @@ type FSharpImplTreeBuilder(lexer, document, decls, lifetime, path, projectedOffs
 
             | _ ->
 
+            let letMark =
+                match bindings with
+                | SynBinding(xmlDoc = XmlDoc xmlDoc) :: _ ->
+                    x.AdvanceToXmlDocOwner(xmlDoc, null, range)
+                    x.Mark()
+                | _ ->
+                    x.Mark()
```
auduchinok, line 104: Why not put `x.Mark()` outside the `match` expression, so it's not duplicated?

ReSharper.FSharp/src/FSharp.Psi.Features/src/Parsing/FSharpImplTreeBuilder.fs
```diff
@@ -74,20 +74,17 @@ type FSharpImplTreeBuilder(lexer, document, decls, lifetime, path, projectedOffs
...
+            | [SynBinding(attributes = attrs; returnInfo = returnInfo; expr = expr; xmlDoc = XmlDoc xmlDoc)] when
                     Range.equals range expr.Range ->
 
+                let letMark = x.MarkOrMarkXmlDocOwner(xmlDoc, range)
                 x.ProcessOuterAttrs(attrs, range)
                 x.AdvanceToStart(range)
                 Assertion.Assert(x.TokenType == FSharpTokenType.EXTERN, "Expecting EXTERN, got: {0}", x.TokenType)
-                let attrs = x.SkipOuterAttrs(attrs, range)
-                x.ProcessAttributeLists(attrs)
+                let _ = x.SkipOuterAttrs(attrs, range)
```
auduchinok, line 87: What exactly does this change do?

ReSharper.FSharp/src/FSharp.Psi.Features/src/Parsing/FSharpTreeBuilderBase.fs
```diff
@@ -117,32 +117,24 @@ type FSharpTreeBuilderBase(lexer, document: IDocument, lifetime, path: VirtualFi
...
-            x.AdvanceToTokenOrRangeStart(expectedType, declarationRange)
-        else
-            x.AdvanceToXmlDoc(xmlDoc, declarationStart)
+        x.AdvanceToXmlDocOwner(xmlDoc, expectedType, declarationRange)
+        let mark = x.Mark()
+        if not xmlDoc.IsEmpty then x.MarkAndDone(xmlDoc.Range, ElementType.XML_DOC_BLOCK)
+        mark
 
-        x.Mark()
+    member x.MarkOrMarkXmlDocOwner(xmlDoc: XmlDoc, declarationRange: range) =
```
auduchinok, line 135: Is this a typo in the name?

ReSharper.FSharp/src/FSharp.Psi.Features/src/Parsing/FSharpImplTreeBuilder.fs
```diff
@@ -230,8 +235,15 @@ type FSharpImplTreeBuilder(lexer, document, decls, lifetime, path, projectedOffs
             match x.ContinueMemberDecl(typeMember.Range) with
             | ValueSome(mark) -> mark
             | _ ->
-                let mark = x.MarkXmlDocOwner(typeMember.XmlDoc, null, typeMember.Range)
-                x.ProcessAttributeLists(typeMember.Attributes)
+                let mark =
+                    match typeMember with
+                    | SynMemberDefn.LetBindings _ ->
+                        x.AdvanceToXmlDocOwner(typeMember.XmlDoc, null, typeMember.Range)
+                        x.Mark()
```
auduchinok, line 242: Why do you call separate `Advance*` and `Mark` method instead of single `MarkXmlDocOwner`, as prior the change? Is there something that got broken in `MarkXmlDocOwner` in this PR?

ReSharper.FSharp/src/FSharp.Psi.Features/src/Parsing/FSharpImplTreeBuilder.fs
```diff
@@ -648,10 +662,9 @@ type FSharpImplTreeBuilder(lexer, document, decls, lifetime, path, projectedOffs
         | [] -> ()
         | binding :: rest ->
 
-        x.ProcessTopLevelBinding(binding)
+        x.ProcessTopLevelBinding(binding, false)
         for binding in rest do
-            x.AdvanceToTokenOrPos(FSharpTokenType.AND, binding.StartPos)
-            x.ProcessTopLevelBinding(binding)
+            x.ProcessTopLevelBinding(binding, true)
```
auduchinok, line 667: The `isSecond` parameter is wrongly `true` for third and subsequent bindings.

ReSharper.FSharp/src/FSharp.Psi.Features/src/Parsing/FSharpImplTreeBuilder.fs
```diff
@@ -625,10 +637,12 @@ type FSharpImplTreeBuilder(lexer, document, decls, lifetime, path, projectedOffs
             if Position.posGt r.End outerRange.Start then attrs else
             x.SkipOuterAttrs(rest, outerRange)
 
-    member x.ProcessTopLevelBinding(binding) =
-        let (SynBinding(_, kind, _, _, attrs, _, _ , headPat, returnInfo, expr, _, _)) = binding
+    member x.ProcessTopLevelBinding(binding, isSecond) =
+        let (SynBinding(_, kind, _, _, attrs, XmlDoc xmlDoc, _ , headPat, returnInfo, expr, range, _)) = binding
 
-        let mark = x.Mark()
+        if isSecond then x.AdvanceToTokenOrRangeStart(FSharpTokenType.AND, if xmlDoc.IsEmpty then range else xmlDoc.Range)
```
auduchinok, line 643: Isn't `if xmlDoc.IsEmpty then range else xmlDoc.Range` logic extracted to the util in other places? Why place it inline here? Is it needed at all given `x.MarkOrMarkXmlDocOwner` is called below?

ReSharper.FSharp/src/FSharp.Psi.Features/src/Parsing/FSharpTreeBuilderBase.fs
```diff
@@ -117,32 +117,24 @@ type FSharpTreeBuilderBase(lexer, document: IDocument, lifetime, path: VirtualFi
...
+        x.AdvanceToTokenOrRangeStart(expectedType, range)
 
     member x.AdvanceToTokenOrPos(tokenType: TokenNodeType, pos: pos) =
         let maxOffset = x.GetOffset(pos)
         while x.CurrentOffset < maxOffset && x.TokenType != tokenType && not x.Eof do
             x.AdvanceLexer()
 
-    // todo: wrong ranges (e.g. missing union case bar)
-    // todo wrong order:
-    // todo multiple blocks:
```
auduchinok, line 135: The multiple blocks todo is removed but I don't see any test case covering it. Was it removed by mistake? I'd expect this todo to be fixed in this PR.

ReSharper.FSharp/src/FSharp.Psi.Features/src/Parsing/FSharpTreeBuilderBase.fs
```diff
@@ -394,33 +384,38 @@ type FSharpTreeBuilderBase(lexer, document: IDocument, lifetime, path: VirtualFi
...
+                    x.Mark()
+
                 for field in fields do
                     x.ProcessField field ElementType.RECORD_FIELD_DECLARATION
                 x.Done(lastFieldRange, fieldListMark, ElementType.RECORD_FIELD_DECLARATION_LIST)
 
             x.Done(range, representationMark, ElementType.RECORD_REPRESENTATION)
 
-        | SynTypeDefnSimpleRepr.Enum(cases, range) ->
+        | SynTypeDefnSimpleRepr.Enum(cases, range) ->
```
auduchinok, line 400: What does this change do?

ReSharper.FSharp/src/FSharp.Psi.Features/src/Parsing/FSharpTreeBuilderBase.fs
```diff
@@ -394,33 +384,38 @@ type FSharpTreeBuilderBase(lexer, document: IDocument, lifetime, path: VirtualFi
...
 
             if not fields.IsEmpty then
-                let (SynField(range = firstFieldRange)) as firstField = fields.Head
+                let SynField(range = firstFieldRange) as firstField = fields.Head
                 let (SynField(range = lastFieldRange)) = List.last fields
 
-                let fieldListMark = x.MarkXmlDocOwner(firstField.XmlDoc, null, firstFieldRange)
+                let fieldListMark =
+                    x.AdvanceToXmlDocOwner(firstField.XmlDoc, null, firstFieldRange)
+                    x.Mark()
```
auduchinok, line 392: Same as in some of the previous comments. Why not `x.MarkXmlDocOwner`?

auduchinok (changes requested):
The multiple blocks case (one of the most important things to do right here) is not covered at all, while a todo comment about it is removed. Why?

Some changes look wrong (like filtering attributes and not processing them).

Some changes don't seem to be needed (e.g. replacing calls to `MarkXmlDocOwner` with `Advance*` and `Mark` direct calls), or there's something got wrong with `MarkXmlDocOwner`.

On top of that there're multiple issues with names like `isSecond` parameter name seems to not cover the second binding case only.

ReSharper.FSharp/src/FSharp.Psi.Features/src/Parsing/FSharpTreeBuilderBase.fs
DedSec256: will fix

ReSharper.FSharp/src/FSharp.Psi.Features/src/Parsing/FSharpTreeBuilderBase.fs
DedSec256: We already have some tests that cover this case:

[One](https://github.com/JetBrains/resharper-fsharp/blob/9321ca1ccbc4d356ebb2737ce79d800c0bd9e688/ReSharper.FSharp/test/data/parsing/Module%20-%20Top%20level%20-%20XmlDoc%2003%20-%20Multiple%20blocks.fs.gold), [two](https://github.com/JetBrains/resharper-fsharp/blob/9321ca1ccbc4d356ebb2737ce79d800c0bd9e688/ReSharper.FSharp/test/data/parsing/Type%20decl%20-%20XmlDoc%2004%20-%20Multiple%20blocks.fs.gold), [three](https://github.com/JetBrains/resharper-fsharp/blob/9321ca1ccbc4d356ebb2737ce79d800c0bd9e688/ReSharper.FSharp/test/data/parsing/Module%20-%20Nested%20-%20XmlDoc%2002.fs.gold)

ReSharper.FSharp/src/FSharp.Psi.Features/src/Parsing/FSharpImplTreeBuilder.fs
DedSec256: yes it should be called isSecondary

ReSharper.FSharp/src/FSharp.Psi.Features/src/Parsing/FSharpImplTreeBuilder.fs
DedSec256: The MarkXmlDocOwner logic, in particular, marks the documentation itself and binds it to the declaration, which is not always required. In some cases, you just need to reach a certain group of declarations without marking the first declaration XmlDoc.

auduchinok: I see. It would probably be better to either add another helper method or add a parameter to `MarkXmlDocOwner` to alter its behaviour. With the current logic a comment about marking the declarations/bindings group would really help.

ReSharper.FSharp/src/FSharp.Psi/src/FSharp.psi
DedSec256: Generally speaking, we should make this block not in grammar at all

ReSharper.FSharp/src/FSharp.Psi.Features/src/Parsing/FSharpTreeBuilderBase.fs
auduchinok: Thanks. Would be great to have something like this too:
```
/// Xml 1
// Comment
/// Xml 2
```

The other existing cases aren't parsed as xml comments for the declarations anymore. Since you've implemented a better logic in the parser, can we rename the tests, so they aren't about multiple blocks anymore?

ReSharper.FSharp/src/FSharp.Psi.Features/src/Parsing/FSharpTreeBuilderBase.fs
DedSec256: I should leave an explanatory comment in the code. 

In some cases, it is necessary that in the absence of XmlDoc, x.Mark is called right on the spot so that specific nodes get into the declaration. For example, in the case of a primary constructor where an access modifier should be included in its declaration.

ReSharper.FSharp/src/FSharp.Psi.Features/src/Parsing/FSharpImplTreeBuilder.fs
DedSec256: I will probably need to revisit this logic again, try to simplify it, and come back to this question later.

ReSharper.FSharp/src/FSharp.Psi.Features/src/Parsing/FSharpImplTreeBuilder.fs
DedSec256: overlooked, this line is also not required

ReSharper.FSharp/src/FSharp.Psi.Features/src/Parsing/FSharpImplTreeBuilder.fs
DedSec256: > Isn't if xmlDoc.IsEmpty then range else xmlDoc.Range logic extracted to the util in other places?

yes we have `xmlDocOwnerStartRange` for this

> Is it needed at all given x.MarkOrMarkXmlDocOwner is called below? It takes xml range into account.

It seems that all the same, it is necessary. `and` should be included in the binding declaration. Since the XmlDoc can appear either before `and` or after, and we only have a range for the XmlDoc and for the declaration without `and`, we must first advance to to XmlDoc or `and`.

DedSec256: I think this also requires a comment in the code

ReSharper.FSharp/src/FSharp.Psi.Features/src/Parsing/FSharpTreeBuilderBase.fs
```diff
@@ -271,29 +261,13 @@ type FSharpTreeBuilderBase(lexer, document: IDocument, lifetime, path: VirtualFi
         if mark.IsSome then
             x.Done(mark.Value, elementType)
 
-    member x.MarkAndProcessAttributesOrIdOrRange(attrs: SynAttributes, xmlDoc: XmlDoc, id: Ident option, range: range) =
+    /// Process xmlDoc and attrbiutes
+    member x.MarkAndProcessIntro(attrs: SynAttributes, xmlDoc: XmlDoc, tokenType: TokenNodeType, range: range) =
+        let mark = x.MarkXmlDocOwner(xmlDoc, tokenType, range)
         match attrs with
```
auduchinok, line 267: Is this `match` actually required after these changes?

ReSharper.FSharp/src/FSharp.Psi.Features/src/Parsing/FSharpTreeBuilderBase.fs
```diff
@@ -117,32 +117,21 @@ type FSharpTreeBuilderBase(lexer, document: IDocument, lifetime, path: VirtualFi
...
-            x.AdvanceToXmlDoc(xmlDoc, declarationStart)
+        x.AdvanceToTokenOrRangeStart(expectedType, declarationRange)
+        let mark = x.Mark()
+        if not xmlDoc.IsEmpty then
+            x.MarkAndDone(xmlDoc.Range, FSharpTokenType.XML_DOC_BLOCK)
+        mark
 
-        x.Mark()
+    member x.Mark(xmlDoc: XmlDoc, declarationRange: range) =
+        if xmlDoc.IsEmpty then x.Mark()
```
auduchinok, line 133: Is this expected that this code must only be called when builder is at the declaration range start position?
Otherwise it may mark some random place prior to the declaration range if `xmlDoc` is empty.

We don't have this requirement in other `Mark*` methods that take ranges to mark, they advance the builder internally.

ReSharper.FSharp/src/FSharp.Psi.Features/src/Parsing/FSharpImplTreeBuilder.fs
```diff
@@ -624,10 +630,12 @@ type FSharpImplTreeBuilder(lexer, document, decls, lifetime, path, projectedOffs
             if Position.posGt r.End outerRange.Start then attrs else
             x.SkipOuterAttrs(rest, outerRange)
 
-    member x.ProcessTopLevelBinding(binding) =
-        let (SynBinding(_, kind, _, _, attrs, _, _ , headPat, returnInfo, expr, _, _, _)) = binding
+    member x.ProcessTopLevelBinding(binding, isSecondary) =
+        let (SynBinding(_, kind, _, _, attrs, XmlDoc xmlDoc, _ , headPat, returnInfo, expr, range, _, _)) = binding
 
-        let mark = x.Mark()
+        if isSecondary then x.AdvanceToTokenOrRangeStart(FSharpTokenType.AND, range)
```
auduchinok, line 636: Do these cases work correctly?
```
let rec f x = ()

/// Yo
and g x = ()
```

```
let rec f x = ()

and g x = ()
```

If yes, is checking the token type needed, or checking the range is sufficient?

ReSharper.FSharp/src/FSharp.Psi.Features/src/Parsing/FSharpTreeBuilderBase.fs
```diff
@@ -74,7 +74,7 @@ type FSharpTreeBuilderBase(lexer, document: IDocument, lifetime, path: VirtualFi
         x.Mark()
 
     member x.Mark() =
-        /// The base member is protected and cannot be used in closures.
+        // The base member is protected and cannot be used in closures.
```
auduchinok, line 77: I'm sorry, but could you please not fix this warning in this PR? 🙂
Let's think about having a quick fix for it instead, having this as a reminder.

ReSharper.FSharp/src/FSharp.Psi.Features/src/Parsing/FSharpTreeBuilderBase.fs
```diff
@@ -400,33 +370,26 @@ type FSharpTreeBuilderBase(lexer, document: IDocument, lifetime, path: VirtualFi
             let representationMark = x.Mark(range)
 
             if not fields.IsEmpty then
-                let (SynField(range = firstFieldRange)) as firstField = fields.Head
```
auduchinok, line 403: Likewise, please don't fix this warning. 🙂
I'll likely allow this case in the analyzer.

ReSharper.FSharp/src/FSharp.Psi/src/Impl/Tree/FSharpComment.cs
```diff
@@ -92,4 +92,17 @@ public DocComment([NotNull] TokenNodeType nodeType, [NotNull] string text) : bas
...
     }
   }
+
+  public class XmlDocBlock : FSharpCompositeElement
+  {
+    [NotNull] private readonly CompositeNodeType myNodeType;
+
+    public XmlDocBlock([NotNull] CompositeNodeType nodeType)
+    {
+      myNodeType = nodeType;
```
auduchinok, line 102: Maybe we should just return `XML_DOC_BLOCK` singleton instead of storing references to it.

auduchinok (commented):
Very nice to see lots of things removed after the FCS changes. Good job on making ranges more consistent there!

There're a few cases where further cleanup/simplification seems possible, though, so please take a look.

ReSharper.FSharp/src/FSharp.Psi/src/Impl/Tree/FSharpComment.cs
auduchinok: JFYI I've just checked and found that other `IDocCommentBlock` implementations do exactly that.

ReSharper.FSharp/src/FSharp.Psi.Features/src/Parsing/FSharpImplTreeBuilder.fs
DedSec256: > Do these cases work correctly?

Yes, there are tests for that:

`Let - Rec 02 - And `
`Let - Top 03 - XmlDoc 01`

> If yes, is checking the token type needed, or checking the range is sufficient?

Yes, still required since `and` is not in binding range

auduchinok: Hm, it looks like I've been looking at `debugPoint` field when writing the comment: its range does include the keyword but I haven't checked if we can use this field reliably.

https://fsprojects.github.io/fantomas-tools/#/ast?data=N4KABGBEAmCmBmBLAdrAzpAXFSAacUiaAYmolmPAIYA2as%2BEkAxgPZwWQ2wAuYATrGaUwADzABeMAAoAlAB1kiqsmhgA5mMkzZkEAF8gA

https://fsprojects.github.io/fantomas-tools/#/ast?data=N4KABGBEAmCmBmBLAdrAzpAXFSAacUiaAYmolmPAIYA2as%2BEkAxgPZwWQ2wAuYATrGaUwADzABeMAAoAlAB1kigPSqwATVaKqyaGADmYyTNmQQAXyA

ReSharper.FSharp/src/FSharp.Psi.Features/src/Parsing/FSharpTreeBuilderBase.fs
```diff
@@ -236,15 +225,16 @@ type FSharpTreeBuilderBase(lexer, document: IDocument, lifetime, path: VirtualFi
...
-                        match moduleKind with
-                        | SynModuleOrNamespaceKind.NamedModule -> FSharpTokenType.MODULE
-                        | SynModuleOrNamespaceKind.DeclaredNamespace -> FSharpTokenType.NAMESPACE
-                        | _ -> null
-                    x.MarkXmlDocOwner(xmlDoc, keywordTokenType, idRange)
+                    match moduleKind with
+                    | SynModuleOrNamespaceKind.NamedModule ->
+                        x.MarkXmlDocOwner(xmlDoc, null, range)
+                    | SynModuleOrNamespaceKind.DeclaredNamespace ->
+                        x.MarkXmlDocOwner(xmlDoc, FSharpTokenType.NAMESPACE, idRange)
```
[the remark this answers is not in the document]
DedSec256: > Can a namespace actually have xml docs?

Yes, they are allowed, but not used in any way

image

> The previous code used to deal with FCS range inconsistencies, do we still need this logic now?

It looks like this piece of code can be simplified, yes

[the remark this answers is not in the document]
auduchinok: > Yes, they are allowed, but not used in any way

Oh, that's sad. It seems we've overlooked this previously, but C# [doesn't allow](https://sharplab.io/#v2:CYLg1APg9DAECaB7AsAKAHYEMC2BTAzgA6YDGusAcvmgN5oC+QA=) it:
Screenshot 2022-02-24 at 16 45 36

ReSharper.FSharp/src/FSharp.Psi.Features/src/Parsing/FSharpTreeBuilderBase.fs
DedSec256: This method is currently used only inside the processing of (top/local) bindings, where it is required that various keywords get inside. Perhaps we should inline its body to the call place.

auduchinok (approved):
Thanks a lot!

PR #322, opened 2021-11-08
Fantomas: add ZoneMarker

DedSec256:
Fix for 

image

### DIFF
--- a/ReSharper.FSharp/src/FSharp.Fantomas.Protocol/src/ZoneMarker.cs
+++ b/ReSharper.FSharp/src/FSharp.Fantomas.Protocol/src/ZoneMarker.cs
@@ -1,0 +1,10 @@
+﻿using JetBrains.Application.BuildScript.Application.Zones;
+using JetBrains.ProjectModel;
+
+namespace JetBrains.ReSharper.Plugins.FSharp.Fantomas.Protocol
+{
+  [ZoneMarker]
+  public class ZoneMarker : IRequire<IProjectModelZone>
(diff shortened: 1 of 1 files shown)


auduchinok:
@DedSec256 What are implications of this change? You're suggesting to merge it to 213 quite late in the release cycle, so it would be good to better understand what's going on. What didn't work without this change? If everything worked as expected, is the error correct at all?

DedSec256:
> What didn't work without this change? If everything worked as expected, is the error correct at all?

Fantomas works as expected. Perhaps for some reason the analyzer did not work correctly, however I also found that the PR error is fixed by fixing the namespace in ProjectModelBase ZoneMarker

image

auduchinok:
> however I also found that the PR error is fixed by fixing the namespace in ProjectModelBase ZoneMarker

What does "fixing the namespace" mean? What was needing fixing?

> Perhaps for some reason the analyzer did not work correctly

If you found it indeed works incorrectly, please do report it.

DedSec256:
> What does "fixing the namespace" mean? What was needing fixing?

Changing namespace to the corresponding location, as shown in the screenshot above
`JetBrains.ReSharper.Plugins.FSharp.ProjectModelBase -> JetBrains.ReSharper.Plugins.FSharp`

> If you found it indeed works incorrectly, please do report it.

It was to the question
>If everything worked as expected, is the error correct at all?

It is possible that the zone analyzer didn't correctly report the error. This should be dealt with.

PR #367, opened 2022-05-04
Initial attempt at ReplaceReturnTypeFix.

nojaf:
Hello @auduchinok, this is a first attempt to add a quick fix action for the `Type constraint mismatch` error when a defined return type differs from the actual type of an expression.

ReplaceReturnType

Please let me know what the best next steps could be here.
I imagine unit tests and edge cases.


ReSharper.FSharp/src/FSharp.Psi.Intentions/src/QuickFixes/ReplaceReturnTypeFix.fs
```diff
@@ -0,0 +1,85 @@
...
+
+type ReplaceReturnTypeFix(expr: IFSharpExpression, fcsErrorMessage: string, replaceFirstType: bool) =
+    inherit FSharpQuickFixBase()
+    
+    let rec visitParent (expr: IFSharpExpression): IFSharpExpression option =
+        match expr.Parent with
+        | :? ISequentialExpr as seqExpr ->
+            if seqExpr.Expressions.Last() <> expr then None else
+            visitParent seqExpr
+        | :? IMatchClause as clause ->
```
auduchinok, line 19: Getting the match clause parent like this is not good here. Is it the `when` expression or is it the expression after `->`? We're only interested in the last one here.
There're generated methods that help with such navigation, take a look at `MatchClauseNavigator` and similar classes.

auduchinok (commented):
@nojaf Could you try to unify the tree traversal with the logic in `RecordCtorReference.GetFcsSymbol`?
https://github.com/JetBrains/resharper-fsharp/blob/net222/ReSharper.FSharp/src/FSharp.Psi/src/Util/FSharpExpressionUtil.cs would be a good place to put it to.

nojaf:
@auduchinok, I'm slowly getting the hang of this.
Let me know where you still see gaps.
I'm sure there are some more cases regarding `ITypeUsage`.

ReSharper.FSharp/src/FSharp.Psi/src/Util/FSharpExpressionUtil.cs
```diff
@@ -35,5 +35,106 @@ public static class FSharpExpressionUtil
...
+          case ILiteralExpr literalExpr:
+            currentExpr = literalExpr;
+            parent = literalExpr.Parent;
+            break;
+          case IMatchClause matchClause:
+            if (matchClause.Expression == currentExpr)
+            {
+              currentExpr = matchClause;
+              parent = matchClause.Parent;
+            }
```
auduchinok, line 59: This looks a bit like reimplementation of `MatchClauseNavigator.GetByExpression` and similar methods. For going up the tree they are preferred way in many cases, since they allow skipping intermediate nodes (when the rule allows) and make it easier to change the tree structure and grammar.

With multi-step navigators used, so it could be done in a way like this:
```csharp
while (true)
{
  if (MatchExprNavigator.GetByClauseExpression(currentExpr) is { } matchExpr)
  {
    currentExpr = matchExpr;
    continue;
  }

  if (SequentialExprNavigator.GetByExpression(currentExpr) is { } seqExpr &&
      seqExpr.ExpressionsEnumerable.Last() == currentExpr)
  {
    currentExpr = seqExpr;
    continue;
  }

  // other navigators go 

  return currentExpr;
}
```

This way you don't have to deal with non-expr parent nodes (like the chameleon expr case above). This is also arguably clearer than a big switch statement. 🙂

Getting parent multiple times is not a problem performance-wise here, as it's a simple field lookup and we only run this code when a caret is at a matching error highlighting type.

ReSharper.FSharp/src/FSharp.Psi/src/Util/FSharpExpressionUtil.cs
```diff
@@ -35,5 +35,106 @@ public static class FSharpExpressionUtil
...
+              currentExpr = null;
+              parent = null;
+            }
+            break;
+          case IIfThenElseExpr ifThenElseExpr:
+            if (ifThenElseExpr.ThenExpr == currentExpr || ifThenElseExpr.ElseExpr == currentExpr)
+            {
+              currentExpr = ifThenElseExpr;
+              parent = ifThenElseExpr.Parent;
+            }
```
auduchinok, line 83: Please note that the overall type of an `if` expression depends on a presence of an `else` branch:
```fsharp
if true then 1 // the overall type is `unit`
if true then 1 else 2 // the overall type is `int`
```

ReSharper.FSharp/src/FSharp.Psi/src/Util/FSharpExpressionUtil.cs
```diff
@@ -35,5 +35,106 @@ public static class FSharpExpressionUtil
...
+            else
+            {
+              currentExpr = null;
+              parent = null;
+            }
+            break;
+          case IForExpr forExpr:
+            if (forExpr.DoExpression == currentExpr)
```
auduchinok, line 91: Likewise, `for` and `while` expressions are always considered `unit`, so the `do` expression doesn't influence the overall type and we can't go up here.

ReSharper.FSharp/test/data/features/quickFixes/replaceReturnType/Replace type with attribute.fs
```diff
@@ -0,0 +1,4 @@
+type FooAttribute() =
+    inherit System.Attribute()
```
auduchinok, line 2: Minor: maybe it could use an existing attribute instead of defining one to keep the test simpler.

ReSharper.FSharp/src/FSharp.Psi.Intentions/src/QuickFixes/ReplaceReturnTypeFix.fs
```diff
@@ -0,0 +1,75 @@
...
+           && isNotNull binding.ReturnTypeInfo.ReturnType then
+            let factory = binding.CreateElementFactory()
+            let typeUsage = factory.CreateTypeUsage(replacementTypeName)
+                
+            match binding.ReturnTypeInfo.ReturnType with
+            | :? INamedTypeUsage as ntu ->
+                ModificationUtil.ReplaceChild(ntu, typeUsage) |> ignore
+            | :? IFunctionTypeUsage as ftu ->
+                ftu.SetReturnTypeUsage(typeUsage) |> ignore
+            | _ -> ()
```
auduchinok, line 75: This doesn't cover parens:
```fsharp
let a: (int) = ""
let b: (int * int) = "", 1
```

auduchinok (commented):
@nojaf The overall approach is much better now, I left a few comments here and there.

ReSharper.FSharp/src/FSharp.Psi/src/Util/FSharpExpressionUtil.cs
```diff
@@ -35,5 +35,106 @@ public static class FSharpExpressionUtil
...
+            else
+            {
+              currentExpr = null;
+              parent = null;
+            }
+            break;
+          case IFSharpExpression fSharpExpression:
+            currentExpr = fSharpExpression;
+            parent = fSharpExpression.Parent;
+            break;
```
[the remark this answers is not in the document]
nojaf: It actually will, as I'm ignoring the `IRecordFieldBinding` parent in the default case.
I'll add a negative test for it.

ReSharper.FSharp/src/FSharp.Psi/src/Util/FSharpExpressionUtil.cs
```diff
@@ -35,5 +35,106 @@ public static class FSharpExpressionUtil
...
+      while (parent != null)
+      {
+        switch (parent)
+        {
+          case IChameleonExpression:
+            return currentExpr as IFSharpExpression;
+          case ILiteralExpr literalExpr:
+            currentExpr = literalExpr;
+            parent = literalExpr.Parent;
+            break;
```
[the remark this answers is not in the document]
nojaf: Yes, this doesn't make sense indeed.
I'll remove it.

ReSharper.FSharp/src/FSharp.Psi/src/Util/FSharpExpressionUtil.cs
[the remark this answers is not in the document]
auduchinok: Yes, right, indeed. Probably no much need for the test, there are many similar cases throughout the grammar, but it's up to you.

ReSharper.FSharp/src/FSharp.Psi/src/Util/FSharpExpressionUtil.cs
```diff
@@ -35,5 +36,54 @@ public static class FSharpExpressionUtil
...
+        }
+        
+        if (SequentialExprNavigator.GetByExpression(currentExpr) is { } seqExpr &&
+            seqExpr.ExpressionsEnumerable.Last() == currentExpr)
+        {
+          currentExpr = seqExpr;
+          continue;
+        }
+
+        if (IfExprNavigator.GetByConditionExpr(currentExpr) is { } ifExpr &&
```
auduchinok, line 59:
```fsharp
let _: int = if true then 1 else "" // the error is in the else branch
let _: int = if 1 then 1 else 2 // we should't go up here
```

ReSharper.FSharp/src/FSharp.Psi/src/Util/FSharpExpressionUtil.cs
```diff
@@ -35,5 +36,54 @@ public static class FSharpExpressionUtil
...
+            ifExpr.ElseExpr != null)
+        {
+          currentExpr = ifExpr;
+          continue;
+        }
+
+        if (BinaryAppExprNavigator.GetByArgument(currentExpr) is { } binaryAppExpr)
```
auduchinok, line 66: We likely need to be more cautious here. Consider this:
```fsharp
let _: int = 1 + ""
```

I'd also check if the logic is the same for piping. For some reason I think there might be some quirks with binary operators and type conversions.

In general, it's usually better to not provide quick fix in cases were we aren't sure we work correctly than breaking the user code in an unexpected way. It's better to gradually enable cases later than to ship it as soon as possible and to deal with a situation where you can't easily fix all the corner cases and don't want to disable the feature too.

ReSharper.FSharp/src/FSharp.Psi/src/Util/FSharpExpressionUtil.cs
```diff
@@ -35,5 +36,54 @@ public static class FSharpExpressionUtil
...
+        }
+
+        if ((TryWithExprNavigator.GetByTryExpression(currentExpr) ??
+             TryWithExprNavigator.GetByClauseExpression(currentExpr)) is { } tryWithExpr)
+        {
+          currentExpr = tryWithExpr;
+          continue;
+        }
+
+        return currentExpr;
```
auduchinok, line 85: The previous implementation also had a `let` expressions part.
```fsharp
let x: int =
    let _ = 1
    ""
```

ReSharper.FSharp/test/src/FSharp.Intentions.Tests/src/QuickFixes/ReplaceReturnTypeTest.fs
```diff
@@ -0,0 +1,28 @@
...
+
+    override x.RelativeTestDataPath = "features/quickFixes/replaceReturnType"
+
+    [<Test>] member x.``Replace type in Constraint Mismatch``() = x.DoNamedTest()
+    [<Test>] member x.``Replace type in infix application``() = x.DoNamedTest()
+    [<Test>] member x.``Replace type in Sequential``() = x.DoNamedTest()
+    [<Test>] member x.``Replace type in MatchClause``() = x.DoNamedTest()
+    [<Test>] member x.``Replace type with attribute``() = x.DoNamedTest()
+    [<Test>] member x.``Replace type in TryWith``() = x.DoNamedTest()
+    [<Test>] member x.``Replace type in MatchLambda``() = x.DoNamedTest()
```
auduchinok, line 19: I'd probably drop `Replace type` in all tests, as this is covered by the class name.

I also tend to use a simple name with a number at the end, as often another related test case is added later and it's easier to group them by using a similar name.

ReSharper.FSharp/test/src/FSharp.Intentions.Tests/src/QuickFixes/ReplaceReturnTypeTest.fs
```diff
@@ -0,0 +1,28 @@
...
+    [<Test; NoHighlightingFound>] member x.``Replace type - Negative 01``() = x.DoNamedTest()
+    [<Test; NotAvailable>] member x.``Replace type - Negative 02``() = x.DoNamedTest()
+    [<Test; NotAvailable>] member x.``Replace type - Negative 03``() = x.DoNamedTest()
+    [<Test; NotAvailable>] member x.``Replace type - Negative 04``() = x.DoNamedTest()
+    [<Test; NoHighlightingFound>] member x.``Replace type - Negative 05``() = x.DoNamedTest()
+    [<Test; NotAvailable>] member x.``Replace type - Negative 06``() = x.DoNamedTest()
+    [<Test; NotAvailable>] member x.``Replace type - Negative 07``() = x.DoNamedTest()
+    [<Test; NoHighlightingFound>] member x.``Replace type - Negative 08``() = x.DoNamedTest()
+    [<Test; NotAvailable>] member x.``Replace type - Negative 09``() = x.DoNamedTest()
```
auduchinok, line 28: It may be a bit clearer to group these tests by the expected error (and to add related cases easier later).
```fsharp
[<Test; NoHighlightingFound>] member x.``No highlighting 01``() = x.DoNamedTest()
[<Test; NoHighlightingFound>] member x.``No highlighting 02``() = x.DoNamedTest()
...
[<Test; NotAvailable>] member x.``Not available - Binary app 01``() = x.DoNamedTest()
[<Test; NotAvailable>] member x.``Not available - Typed pattern 01``() = x.DoNamedTest()
...
```

auduchinok (commented):
Here goes another round. 🙂

ReSharper.FSharp/src/FSharp.Psi/src/Util/FSharpExpressionUtil.cs
```diff
@@ -35,5 +36,58 @@ public static class FSharpExpressionUtil
...
+        {
+          currentExpr = lambdaExpr;
+          continue;
+        }
+
+        if (TupleExprNavigator.GetByExpression(currentExpr) is { } tupleExpr)
+        {
+          currentExpr = tupleExpr;
+          continue;
+        }
```
auduchinok, line 81: Could you show an example where we need to go up from a tuple expression?

ReSharper.FSharp/src/FSharp.Psi/src/Util/FSharpExpressionUtil.cs
```diff
@@ -35,5 +36,58 @@ public static class FSharpExpressionUtil
...
+        {
+          currentExpr = letOrUseExpr;
+          continue;
+        }
+
+        if (LambdaExprNavigator.GetByExpression(currentExpr) is { } lambdaExpr)
+        {
+          currentExpr = lambdaExpr;
+          continue;
+        }
```
auduchinok, line 75: Likewise, can we go up from any lambda body?

ReSharper.FSharp/src/FSharp.Psi/src/Util/FSharpExpressionUtil.cs
```diff
@@ -35,5 +36,58 @@ public static class FSharpExpressionUtil
...
+
+        if (TupleExprNavigator.GetByExpression(currentExpr) is { } tupleExpr)
+        {
+          currentExpr = tupleExpr;
+          continue;
+        }
+
+        if (BinaryAppExprNavigator.GetByArgument(currentExpr) is { } binaryAppExpr)
```
auduchinok, line 83:
```fsharp
let x: int = 1 + ""
```

ReSharper.FSharp/src/FSharp.Psi/src/Util/FSharpExpressionUtil.cs
nojaf: See test `Tuple return type.fs`

ReSharper.FSharp/src/FSharp.Psi/src/Util/FSharpExpressionUtil.cs
nojaf: See test `FunctionType 01.fs`.

ReSharper.FSharp/src/FSharp.Psi/src/Util/FSharpExpressionUtil.cs
nojaf: Yes, still on my to-do list.

ReSharper.FSharp/src/FSharp.Psi/src/Util/FSharpExpressionUtil.cs
auduchinok: I see, thanks. We should carefully consider possible syntax and FCS types here:
```fsharp
let f: int -> double -> string = Unchecked.defaultof<_>
let g: int -> int -> string = f
```
```fsharp
let h: int * (int -> int) = 1, fun x -> ""
```

ReSharper.FSharp/src/FSharp.Psi/src/Util/FSharpExpressionUtil.cs
auduchinok: Another interesting case:
```fsharp
let x: int -> int = 1, 1
```

ReSharper.FSharp/src/FSharp.Psi.Services/src/Daemon/Highlightings/FcsErrors.xml
```diff
@@ -525,5 +525,18 @@
     <Behavour overlapResolvePolicy="NONE"/>
     <QuickFix>ReplaceReturnTypeFix</QuickFix>
   </Error>
-  
+
+  <Error staticGroup="FSharpErrors" name="TypeMisMatchTuplesHaveDifferingLengths" ID="FS0001: Type mismatch. Expecting">
```
auduchinok, line 529: Maybe we probably should check this inside the quick fix's `IsAvailable` method instead of creating different error types. I'll need to think about it.

Parsing the error also seem a too ad-hoc and a bit dangerous, it would be much better to work with FCS's `FSharpType` instead. It's not always possible to get that view of the expected type, though.

nojaf:
Would it be an option to put some more logic into the `override this.IsAvailable` member to skip the more complicated scenarios?

ReSharper.FSharp/src/FSharp.Psi.Daemon/src/Stages/FcsErrorsStageProcessBase.fs
```diff
@@ -184,6 +187,17 @@ type FcsErrorsStageProcessBase(fsFile, daemonProcess) =
...
+                let expr = nodeSelectionProvider.GetExpressionInRange(fsFile, range, false, null)
+                TypeDoesNotMatchTypeError(expectedType, actualType, expr, error.Message)
+
+            | Regex ifBranchSatisfyContextTypeRequirements [expectedType; actualType] ->
+                let expr = nodeSelectionProvider.GetExpressionInRange(fsFile, range, false, null)
+                IfExpressionNeedsTypeToSatisfyTypeRequirementsError(expectedType, actualType, expr, error.Message)
+
+            | Regex typeMisMatchTupleLengths [expectedType; actualType] ->
+                let expr = nodeSelectionProvider.GetExpressionInRange(fsFile, range, false, null)
+                TypeMisMatchTuplesHaveDifferingLengthsError(expectedType, actualType, expr, error.Message)
```
auduchinok, line 200: There's something wrong with reported types when item count mismatch:

```fsharp
let x: int * int * int = 1, ""
```

Screenshot 2022-05-24 at 17 23 33

Let's disable this case until the error is fixed?

ReSharper.FSharp/src/FSharp.Psi.Intentions/src/QuickFixes/ReplaceReturnTypeFix.fs
```diff
@@ -0,0 +1,91 @@
...
+        let refPat = binding.HeadPattern.As<IReferencePat>()
+        if isNull refPat then () else
+
+        let symbolUse = refPat.GetFcsSymbolUse()
+        if isNull symbolUse then () else
+
+        if isNotNull binding.ReturnTypeInfo
+           && isNotNull binding.ReturnTypeInfo.ReturnType then
+            let factory = binding.CreateElementFactory()
+            let typeUsage = factory.CreateTypeUsage(replacementTypeName)
```
auduchinok, line 66: I consider this creation of a type usage from a text parsed in an error text to be a bit dangerous but we can try to use it for the time being.

ReSharper.FSharp/src/FSharp.Psi.Intentions/src/QuickFixes/ReplaceReturnTypeFix.fs
```diff
@@ -0,0 +1,91 @@
...
+                | returnType -> returnType
+
+            match returnType with
+            | :? INamedTypeUsage as ntu ->
+                ModificationUtil.ReplaceChild(ntu, typeUsage) |> ignore
+            | :? IFunctionTypeUsage as ftu ->
+                match mostOuterParentExpr with
+                | :? ITupleExpr ->
+                    ModificationUtil.ReplaceChild(ftu, typeUsage) |> ignore
+                | _ -> ftu.SetReturnTypeUsage(typeUsage) |> ignore
```
auduchinok, line 79: This doesn't handle this case correctly:

```fsharp
let x = 1, 2
let f: int -> int = x 
```

ReSharper.FSharp/src/FSharp.Psi/src/Util/FSharpExpressionUtil.cs
```diff
@@ -35,5 +36,58 @@ public static class FSharpExpressionUtil
 
     public static readonly Func<IFSharpExpression, bool> IsSimpleValueExpressionFunc = IsSimpleValueExpression;
     public static readonly Func<IFSharpExpression, bool> IsLiteralExpressionFunc = IsLiteralExpression;
+
+    // TODO: change name
+    public static IFSharpExpression GetOuterMostParentExpression(this IFSharpExpression expression)
```
auduchinok, line 41: Maybe we use something like `GetOutermostParentExpressionFromItsReturn` for now (though, I find it too long for to be good).

ReSharper.FSharp/src/FSharp.Psi/src/Util/FSharpExpressionUtil.cs
auduchinok: Both these cases don't work correctly with the current implementation. We should disable going up from lambda expressions for now.

ReSharper.FSharp/src/FSharp.Psi/src/Util/FSharpExpressionUtil.cs
auduchinok: We shouldn't go up here, but it's OK to go up in the quick fix (as you already seem to do).

ReSharper.FSharp/src/FSharp.Psi/src/Util/FSharpExpressionUtil.cs
auduchinok: Please remove this `if` for now 🙂

nojaf:
My pleasure, thank you for all the guidance along the way!

PR #369, opened 2022-05-12
Run Fantomas always on .NET Core

DedSec256:
- [x] Check CI


ReSharper.FSharp/src/FSharp.Fantomas.Host/Fantomas.Host.win.runtimeconfig.json
```diff
@@ -0,0 +1,10 @@
+﻿{
+  "runtimeOptions": {
+    "tfm": "net6",
+    "rollForward": "Major",
+    "framework": {
+      "name": "Microsoft.WindowsDesktop.App",
```
ForNeVeR, line 6: _(Minor)_ I wonder: why does it require `Microsoft.WindowsDesktop.App` on Windows?

ForNeVeR (approved):

ReSharper.FSharp/src/FSharp.Fantomas.Host/Fantomas.Host.win.runtimeconfig.json
DedSec256: RD-protocol we use requires System.Windows.Threading.Dispatcher

PR #378, opened 2022-07-05
Generate signature file

nojaf:
Hello @auduchinok, I'd like to expose the [GenerateSignature](https://github.com/dotnet/fsharp/blob/0200797d8d943f0e73d5de50e7bda4f654a96a1e/src/Compiler/Service/FSharpCheckerResults.fsi#L397) functionality from FCS to easily add a signature file for a single file.

This first commit contains some horrible pseudo code to illustrate what I'm after.
The feature:
- Should be executable if there is no signature file present.
- Should generate the signature file (using the FCS).
- Should update the `fsproj` to include the signature file.
- (Should maybe format the generated file, as the output isn't always that pretty).

Would you be ok with having this? If so, do you have any pointers on how to best approach this?


nojaf:
@auduchinok I'm able to add the new signature file before the implementation file.
However, there is a small glitch with the UI:
image

`Library.fsi` is the new file and is shown at the top. It is inserted correctly into the `fsproj` thought. 
Thanks again for the pointers!

PR #380, opened 2022-07-06
Add function to signature file

nojaf:
Hello @auduchinok, I'd like to add a code action that adds a `val` entry to a signature file for a function that is not exposed.

I've made a very raw proposal of what I have in mind. 
The `val` generation itself would not be a string concatenation but you get the idea 😸.

Would you be ok with having this? If so, do you have any pointers on how to best approach this?

### DIFF
--- a/ReSharper.FSharp/src/FSharp.Psi.Intentions/src/Intentions/AddFunctionToSignatureAction.fs
+++ b/ReSharper.FSharp/src/FSharp.Psi.Intentions/src/Intentions/AddFunctionToSignatureAction.fs
@@ -25,7 +25,20 @@ type AddFunctionToSignatureFileAction(dataProvider: FSharpContextActionDataProvi
         | null -> None
         | pat -> pat.GetFcsSymbol().SignatureLocation
 
-    let addChildAfter child anchor = ModificationUtil.AddChildAfter(anchor, child)
+    let addChildAfter (newlineNode: ITreeNode) (insertNewline: bool) (child: ITreeNode) (anchor: ITreeNode) =
+        let anchor =
+            let anchor =
+                if anchor.NodeType <> newlineNode.NodeType then
+                    ModificationUtil.AddChildAfter(anchor, newlineNode)
+                else
+                    anchor
+
+            if insertNewline then
+                ModificationUtil.AddChildAfter(anchor, newlineNode)
+            else
+                anchor
+
+        ModificationUtil.AddChildAfter(ModificationUtil.AddChildAfter(anchor, child), newlineNode)
     
     override x.Text = "Add to signature file"
     override this.IsAvailable _ =
@@ -46,20 +59,33 @@ type AddFunctionToSignatureFileAction(dataProvider: FSharpContextActionDataProvi
 
         match SignatureFile.tryMkBindingSignature letBindings moduleDecl with
         | None -> null
-        | Some (sigDeclNode, openStatements, sigFile) ->
-            let newlineNode = NewLine(sigDeclNode.GetLineEnding()) :> ITreeNode
+        | Some response ->
+            let sigFile = response.SigFile
+            let openStatements = response.OpenStatements
+            let newlineNode = NewLine(response.SigDeclNode.GetLineEnding()) :> ITreeNode
+            let addChildAfter : bool -> ITreeNode -> ITreeNode -> ITreeNode  = addChildAfter newlineNode
+
+            let lastExistingOpenStatement =
+                response.SigModule.Members
+                |> Seq.choose (function | :? IOpenStatement as os -> Some os | _ -> None)
+                |> Seq.tryLast
             
-            sigFile.LastChild
-            |> addChildAfter newlineNode
-            |> fun lastChild ->
-                if List.isEmpty openStatements then
-                    lastChild
-                else
-                    (lastChild, openStatements)
-                    ||> List.fold (fun acc openStatement -> addChildAfter openStatement acc |> addChildAfter newlineNode)
-                    |> addChildAfter newlineNode
-            |> addChildAfter sigDeclNode
-            |> addChildAfter newlineNode
+            let addOpeningStatements node =
+                (node, openStatements)
+                ||> List.fold (fun acc openStatement -> addChildAfter false openStatement acc)
+            
+            let lastNode =
+                match lastExistingOpenStatement, openStatements.IsEmpty with
+                | Some lastOpenStatement, false ->
+                    // TODO: this won't work if there are other vals in the signature file
+                    addOpeningStatements lastOpenStatement
+                | None, false ->
+                    addOpeningStatements (ModificationUtil.AddChildAfter(sigFile.LastChild, newlineNode))
+                | _, true ->
+                    sigFile.LastChild
+
+            lastNode
+            |> addChildAfter true response.SigDeclNode
             |> ignore
 
             null
--- a/ReSharper.FSharp/src/FSharp.Psi.Intentions/src/QuickFixes/UpdateSignatureFileFix.fs
+++ b/ReSharper.FSharp/src/FSharp.Psi.Intentions/src/QuickFixes/UpdateSignatureFileFix.fs
@@ -43,8 +43,9 @@ type UpdateSignatureFileFix(binding: IBinding) =
 
         match SignatureFile.tryMkBindingSignature letBindings moduleDecl with
         | None -> ()
-        | Some (sigDeclNode, _, sigFile) ->
-            
+        | Some response ->
+
+        let ({ SigDeclNode = sigDeclNode }: SignatureFile.SignatureBindingResponse) = response
         let implHeadPat =
             let binding = letBindings.Bindings.FirstOrDefault() in
             if isNull binding then null else
@@ -54,7 +55,7 @@ type UpdateSignatureFileFix(binding: IBinding) =
 
         if isNull implHeadPat then () else
 
-        let mdl = sigFile.ModuleDeclarations.FirstOrDefault()
+        let mdl = response.SigFile.ModuleDeclarations.FirstOrDefault()
         let existingBinding =
             Seq.cast<IBindingSignature> mdl.MembersEnumerable
             |> Seq.tryFind (fun b ->
--- a/ReSharper.FSharp/src/FSharp.Psi.Services/src/Util/SignatureFileUtil.fs
+++ b/ReSharper.FSharp/src/FSharp.Psi.Services/src/Util/SignatureFileUtil.fs
@@ -16,10 +16,19 @@ let private (|DeclaredNameInPattern|_|) (pat:IFSharpPattern) =
         | _ -> None
     | _ -> None
 
+type SignatureBindingResponse =
+    {
+        SigDeclNode: IBindingSignature
+        OpenStatements: IOpenStatement list
+        SigModule: INamedModuleDeclaration
+    }
+    
+    member this.SigFile = this.SigModule.Parent :?> IFSharpSigFile
+
 let tryMkBindingSignature
     (letBindings: ILetBindings)
     (moduleDecl: IModuleDeclaration)
-    : (IBindingSignature *  IOpenStatement list * IFSharpSigFile) option
+    : SignatureBindingResponse option
     =
     if isNull moduleDecl then None else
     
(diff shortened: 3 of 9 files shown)


auduchinok:
@nojaf This is an interesting feature, but limiting it to certain kinds of symbols like module functions doesn't feel right. We should design big features like this properly.

ReSharper.FSharp/src/FSharp.Psi.Features/src/Parsing/FSharpParser.fs
```diff
@@ -62,10 +62,12 @@ type FSharpParser(lexer: ILexer, document: IDocument, path: VirtualFileSystemPat
         let path = if isNotNull sourceFile then sourceFile.GetLocation() else null
         FSharpParser(lexer, document, path, sourceFile, checkerService, symbolsCache)
 
-    new (lexer, document, sourceFile, checkerService, symbolsCache) =
-        FSharpParser(lexer, document, FSharpParser.SandBoxPath, sourceFile, checkerService, symbolsCache)
+    new (lexer, document, sourceFile: IPsiSourceFile, checkerService, symbolsCache) =
+        let path = if sourceFile.Name.EndsWith(".fsi") then FSharpParser.SandBoxSignaturePath else FSharpParser.SandBoxPath
```
auduchinok, line 66: `sourceFile` may be `null` here, it's used as a context in which a new node is being parsed, and we can just use its path.
```suggestion
        let path = if isNotNull sourceFile then sourceFile.GetLocation() else FSharpParser.SandBoxPath
```

nojaf:
Hey @auduchinok, I wrote some code and could probably use a review 😸.
Anyway, I created a quick fix to update an existing signature file if the `val` no longer matches.

UpdateSignatureFile

ReSharper.FSharp/src/FSharp.Psi.Intentions/src/Intentions/AddFunctionToSignatureAction.fs
```diff
@@ -0,0 +1,91 @@
...
+    override this.ExecutePsiTransaction(solution, progress) =
+        use writeCookie = WriteLockCookie.Create(true)
+        use disableFormatter = new DisableCodeFormatter()
+        let letBindings = dataProvider.GetSelectedElement<ILetBindingsDeclaration>()
+        let moduleDecl = ModuleDeclarationNavigator.GetByMember(letBindings)
+
+        match SignatureFile.tryMkBindingSignature letBindings moduleDecl with
+        | None -> null
+        | Some response ->
+            let sigFile = response.SigFile
```
nojaf, line 63: Hey @auduchinok, any advice on handling blank lines?
I'm making a mess in the code here.

I'm faced with scenarios where you want to introduce a blank line (I think).
Like from

```fsharp
module Foo
```

to

```fsharp
module Foo

val bar: int
```

But you don't want this all the time:

Example from
```fsharp
module Foo

open System
```

to

```fsharp
module Foo

open System
open System.Collections.Generic

val bar : IDictionary<string,DateTime>
```

You don't want to add a blank line between the open statements.

ReSharper.FSharp/src/FSharp.Psi.Services/src/Util/SignatureFileUtil.fs
```diff
@@ -0,0 +1,172 @@
...
+            |> String.concat " -> "
+            |> sprintf "(%s)"
+        elif t.IsTupleType then
+            t.GenericArguments
+            |> Seq.map getTypeName
+            |> String.concat " * "
+        else
+            "???"
+
+    let rec visitAccessPath (t: FSharpType) : string list =
```
nojaf, line 95: There are probably better ways to do all this.
I'm all ears if you can give some pointers 😸.

ReSharper.FSharp/src/FSharp.Psi.Intentions/src/QuickFixes/UpdateSignatureFileFix.fs
```diff
@@ -0,0 +1,73 @@
...
+
+        if isNull implHeadPat then () else
+
+        let mdl = response.SigFile.ModuleDeclarations.FirstOrDefault()
+        let existingBinding =
+            Seq.cast<IBindingSignature> mdl.MembersEnumerable
+            |> Seq.tryFind (fun b ->
+                match b.HeadPattern with
+                | :? ITopReferencePat as sigPat ->
+                    sigPat.CompiledName = implHeadPat.CompiledName
```
auduchinok, line 64: Please use source names bindings here, as the compiled names use attributes and could also be out of sync.

ReSharper.FSharp/src/FSharp.Psi.Services/src/Daemon/Highlightings/FcsErrors.xml
```diff
@@ -560,4 +560,20 @@
     <QuickFix>RemoveXmlDocFix</QuickFix>
   </Warning>
 
+  <Error compilerGroup="FSharpErrors" name="SignatureFileMismatch" ID="FS0034: Signature file mismatch">
+    <Parameter type="IFSharpPattern" name="pattern"/>
+    <Message value="Signature file mismatch"/>
```
auduchinok, line 565: This message doesn't explain the user what's wrong. We should use the FCS message in this case.

ReSharper.FSharp/src/FSharp.Psi.Services/src/Daemon/Highlightings/FcsErrors.xml
```diff
@@ -560,4 +560,20 @@
...
+    <Range>pattern.GetHighlightingRange()</Range>
+    <Behavour overlapResolvePolicy="NONE"/>
+    <QuickFix>UpdateSignatureFileFix</QuickFix>
+  </Error>
+  
+  <Warning compilerGroup="FSharpErrors" name="ArgumentNameMismatch" ID=" FS3218: The argument names in the signature and implementation do not match">
+    <Parameter type="IFSharpPattern" name="pattern"/>
+    <Message value="Signature file mismatch"/>
```
auduchinok, line 573: Likewise, please use a more descriptive message.

ReSharper.FSharp/src/FSharp.Psi.Services/src/Util/SignatureFileUtil.fs
```diff
@@ -0,0 +1,172 @@
...
+    member this.SigFile = this.SigModule.Parent :?> IFSharpSigFile
+
+let tryMkBindingSignature
+    (letBindings: ILetBindings)
+    (moduleDecl: IModuleDeclaration)
+    : SignatureBindingResponse option
+    =
+    if isNull moduleDecl then None else
+    
+    let allDecls = moduleDecl.DeclaredElement.GetDeclarations() |> Seq.cast<INamedModuleDeclaration> |> Seq.toArray
```
auduchinok, line 35: This should work for nested modules too.

ReSharper.FSharp/src/FSharp.Psi.Services/src/Util/SignatureFileUtil.fs
```diff
@@ -0,0 +1,172 @@
...
+
+    let sigDecl = Array.find (fun (nmd: INamedModuleDeclaration) ->
+        match nmd.Parent with
+        | :? IFSharpSigFile -> true
+        | _ -> false) allDecls
+
+    let elementFactory = sigDecl.CreateElementFactory()
+    let binding = letBindings.Bindings.First()
+    let refPat = binding.HeadPattern.As<IReferencePat>()
+    let name = refPat.ReferenceName.Identifier.Name
```
auduchinok, line 51: There will likely be other features and cases where `tryMkBindingSignature` is going to be used, so we should not assume it's only called on `IReferencePat` pattern bindings. This will throw NRE.

ReSharper.FSharp/src/FSharp.Psi.Services/src/Util/SignatureFileUtil.fs
```diff
@@ -0,0 +1,172 @@
...
+            |> Seq.choose (function | :? IOpenStatement as os -> Some os | _ -> None)
+            |> Seq.map (fun openStatement -> openStatement.ReferenceName.QualifiedName)
+
+        set
+            [|
+                yield "Microsoft.FSharp.Core"
+                yield "Microsoft.FSharp.Collections"
+                yield sigDecl.CompiledName
+                yield! existingOpenStatements
+            |]
```
auduchinok, line 115: Let's try to use `OpensUtil.OpenedModulesProvider.OpenedModuleScopes` instead? It should properly deal with all implicitly opened modules, take into account the current one declaration and opened `open` statements.

ReSharper.FSharp/src/FSharp.Psi.Services/src/Util/SignatureFileUtil.fs
```diff
@@ -0,0 +1,172 @@
...
+              yield! otherPaths ]
+        |> List.map elementFactory.CreateOpenStatement
+
+    let namedParameters =
+        Seq.zip types parameters
+        |> Seq.map (fun (t, p) ->
+            let typeName = getTypeName t
+            match p.Pattern.IgnoreInnerParens() with
+            | DeclaredNameInPattern name -> $"{name}: {typeName}"
+            | :? ITuplePat as tuplePat ->
```
auduchinok, line 133: This may break with nested parens. We should check if, for example, in this code the two tuples have the same representation in signatures and compiled code or not:
```fsharp
let f (a, b) ((c, d)) = ...
```

ReSharper.FSharp/src/FSharp.Psi.Intentions/src/Intentions/AddFunctionToSignatureAction.fs
```diff
@@ -0,0 +1,91 @@
...
+                |> Seq.tryLast
+            
+            let addOpeningStatements node =
+                (node, openStatements)
+                ||> List.fold (fun acc openStatement -> addChildAfter false openStatement acc)
+            
+            let lastNode =
+                match lastExistingOpenStatement, openStatements.IsEmpty with
+                | Some lastOpenStatement, false ->
+                    // TODO: this won't work if there are other vals in the signature file
```
auduchinok, line 80: This definitely needs to be fixed.

ReSharper.FSharp/src/FSharp.Psi.Intentions/src/QuickFixes/UpdateSignatureFileFix.fs
```diff
@@ -0,0 +1,73 @@
...
+            match tp.Parent with
+            | :? IParenPat as pat ->
+                UpdateSignatureFileFix(BindingNavigator.GetByParameterPattern(pat))
+            | _ -> UpdateSignatureFileFix(Unchecked.defaultof<IBinding>)
+        | _ -> UpdateSignatureFileFix(Unchecked.defaultof<IBinding>)
+    
+    override this.Text = "Update signature file"
+    
+    override this.IsAvailable _ =
+        Option.isSome moduleDecl
```
auduchinok, line 37: We should also check that types used in a new signature are available in the signature file:
```fsharp
module M

type T() = class end

let f (_: T) = ()
```
```fsharp
module M

val f: int -> unit
```

ReSharper.FSharp/src/FSharp.Psi.Services/src/Util/SignatureFileUtil.fs
auduchinok: `RequireQualifiedAccess` and `AutoOpen` attributes may make us want insert a different namespace/module open than the one that a type is defined in. `OpensUtil.getModuleToOpen` takes these things into account.

auduchinok (commented):
@nojaf It's a very good start!

My main concern here is we really need to make the changes produced by quick fixes smaller and not replace the whole member, otherwise it's won't be possible to properly integrate it with other features and would add technical debt as it'd require rewriting these things later. I'd be happy to try to pair-program the changes to make the quick fixes produces smaller deltas if you want.

It also looks like the PR actually contains three big and distinct features (change a parameter name in signature, change a type in signature, and add a module value signature). It might be easier to merge them if all three were done in separate PRs. 🙂

ReSharper.FSharp/src/FSharp.Psi.Intentions/src/Intentions/AddFunctionToSignatureAction.fs
auduchinok: There's some code in `OpensUtil.addOpen` and places like `GenerateOverrides.addEmptyLineAfterIfNeeded`. We could copy or generalize a few pieces from them.

The best approach would be to finally finish the formatter for top level members, and it'd do these things automatically when adding a member node. We could try to enable it here just to see where/if it breaks. 🙂

PR #385, opened 2022-07-27
Fix gradle project model for protocol

mfilippov:


auduchinok (approved):
Wow, it makes editing the models so much nicer, thank you very much!

PR #388, opened 2022-08-05
Psi/FSharpMethodParameter: fix [<ParamArray>] check

DedSec256:
Fix for [RIDER-80565](https://youtrack.jetbrains.com/issue/RIDER-80565/SystemParamArray-is-not-handled-correctly-when-referencing-from-C-code)

Before:
image

After:
image

### DIFF
--- a/ReSharper.FSharp/src/FSharp.Psi/src/Impl/DeclaredElement/FSharpMethodParameter.cs
+++ b/ReSharper.FSharp/src/FSharp.Psi/src/Impl/DeclaredElement/FSharpMethodParameter.cs
@@ -55,7 +55,7 @@ namespace JetBrains.ReSharper.Plugins.FSharp.Psi.Impl.DeclaredElement
     }
 
     public override ParameterKind Kind => FSharpSymbol.MapParameterKind();
-    public override bool IsParameterArray => Owner is IFSharpMethod && FSharpSymbol.IsParamArrayArg;
+    public override bool IsParameterArray => Owner is IFSharpFunction && FSharpSymbol.IsParamArrayArg;
 
     // todo: implement IsCliOptional in FCS
     public override bool IsOptional =>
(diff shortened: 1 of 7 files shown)


ReSharper.FSharp/src/FSharp.Psi/src/Impl/DeclaredElement/FSharpMethodParameter.cs
```diff
@@ -55,7 +55,7 @@ public override DefaultValue GetDefaultValue()
     }
 
     public override ParameterKind Kind => FSharpSymbol.MapParameterKind();
-    public override bool IsParameterArray => Owner is IFSharpMethod && FSharpSymbol.IsParamArrayArg;
+    public override bool IsParameterArray => Owner is IFSharpFunction && FSharpSymbol.IsParamArrayArg;
```
DedSec256, line 58: Possibly we should rename FSharpMethodParameter to FSharpParameter?

DedSec256:
.gold updated to match behavior in C#

image

ReSharper.FSharp/src/FSharp.Psi/src/Impl/DeclaredElement/FSharpMethodParameter.cs
auduchinok: Yes, I think we could do that.

auduchinok (approved):

ReSharper.FSharp/src/FSharp.Psi/src/Impl/DeclaredElement/FSharpMethodParameter.cs
auduchinok: Though, it would have been better if it was a separate commit, as the fix for the param array is a bit hidden by unrelated changes now.

PR #389, opened 2022-08-06
Lambda analyzer: fix for method lambda arguments

DedSec256:
Fix for [RIDER-68269](https://youtrack.jetbrains.com/issue/RIDER-68269/False-positive-Lambda-can-be-replaced-with)

### DIFF
--- a/ReSharper.FSharp/src/FSharp.Psi.Daemon/src/Analyzers/LambdaAnalyzer.fs
+++ b/ReSharper.FSharp/src/FSharp.Psi.Daemon/src/Analyzers/LambdaAnalyzer.fs
@@ -137,7 +137,7 @@ type LambdaAnalyzer() =
                     // If the body of the lambda consists of a method call,
                     // and the method to which the lambda is passed has overloads,
                     // then it cannot be unambiguously determined whether the lambda can be simplified
-                    match getAllMethods ref.FSharpFile reference false "LambdaAnalyzer.getMethods" with
+                    match getAllMethods reference false "LambdaAnalyzer.getMethods" with
                     | None
                     | Some (_, None)
                     | Some (_, Some [])
--- a/ReSharper.FSharp/src/FSharp.Psi.Daemon/src/Analyzers/LambdaAnalyzer.fs
+++ b/ReSharper.FSharp/src/FSharp.Psi.Daemon/src/Analyzers/LambdaAnalyzer.fs
@@ -10,7 +10,6 @@ open JetBrains.ReSharper.Plugins.FSharp.Psi.Features.Util.FSharpMethodInvocation
 open JetBrains.ReSharper.Plugins.FSharp.Psi.Impl
 open JetBrains.ReSharper.Plugins.FSharp.Psi.Tree
 open JetBrains.ReSharper.Plugins.FSharp.Util.FSharpSymbolUtil
-open JetBrains.ReSharper.Psi
 open JetBrains.ReSharper.Psi.ExtensionsAPI
 open JetBrains.ReSharper.Psi.Tree
 open JetBrains.Util
@@ -85,18 +84,15 @@ type LambdaAnalyzer() =
 
         compareArgsRec expr 0 null
 
-    let canBeNamedArg (expr: IBinaryAppExpr) =
+    let isNamedArg (expr: IBinaryAppExpr) =
         match expr.Parent with
         | :? IParenExpr -> false
-        | _ ->
-        if isNull expr.Operator then false else
-        let exprReference = expr.Operator.Reference
-        exprReference.GetName() = "="
+        | _ -> expr.ShortName = "="
 
     let tryCreateWarning (ctor: ILambdaExpr * 'a -> #IHighlighting) (lambda: ILambdaExpr, replacementExpr: 'a as arg) isFSharp6Supported =
         let lambda = lambda.IgnoreParentParens()
         let binaryExpr = BinaryAppExprNavigator.GetByRightArgument(lambda)
-        if isNotNull binaryExpr && not (canBeNamedArg binaryExpr) then ctor arg else
+        if isNotNull binaryExpr && not (isNamedArg binaryExpr) then ctor arg else
         let argExpr = if isNull binaryExpr then lambda else binaryExpr :> _
         let appTuple = TupleExprNavigator.GetByExpression(argExpr)
         let app = getArgsOwner argExpr
--- a/ReSharper.FSharp/src/FSharp.Psi.Daemon/src/Analyzers/LambdaAnalyzer.fs
+++ b/ReSharper.FSharp/src/FSharp.Psi.Daemon/src/Analyzers/LambdaAnalyzer.fs
@@ -10,6 +10,7 @@ open JetBrains.ReSharper.Plugins.FSharp.Psi.Features.Util.FSharpMethodInvocation
 open JetBrains.ReSharper.Plugins.FSharp.Psi.Impl
 open JetBrains.ReSharper.Plugins.FSharp.Psi.Tree
 open JetBrains.ReSharper.Plugins.FSharp.Util.FSharpSymbolUtil
+open JetBrains.ReSharper.Psi
 open JetBrains.ReSharper.Psi.ExtensionsAPI
 open JetBrains.ReSharper.Psi.Tree
 open JetBrains.Util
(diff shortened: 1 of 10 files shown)


ReSharper.FSharp/src/FSharp.Psi.Daemon/src/Analyzers/LambdaAnalyzer.fs
```diff
@@ -84,45 +84,73 @@ type LambdaAnalyzer() =
...
-                not mHasOverload
-            else false
-        | _ -> false
+                
+            // If the lambda is passed instead of a delegate,
+            // then in F# < 6.0 there is almost never an implicit cast for the lambda simplification 
+            if argIsDelegate && not isFSharp6Supported then null else
+
+            match exprToReplace.As<IFSharpExpression>().IgnoreInnerParens() with
+            | :? IReferenceExpr as ref ->
```
auduchinok, line 117: Maybe use something like this to reduce nesting? Would be cool if there was an easy way to split the logic for existing expressions and new expression names, so we don't do these kind of type checks.
```fsharp
let refExpr = exprToReplace.As<IFSharpExpression>().IgnoreInnerParens().As<IReferenceExpr>()
if isNul refExpr then ctor arg else

...
```

If `GetMethods` if refactored out successfully, then this match expression could become small enough to split and to repeat the util function usage.

ReSharper.FSharp/src/FSharp.Psi.Daemon/src/Analyzers/LambdaAnalyzer.fs
```diff
@@ -84,45 +84,73 @@ type LambdaAnalyzer() =
...
+    let getWarningIfApplicable (ctor: ILambdaExpr * 'a -> #IHighlighting) (lambda: ILambdaExpr, exprToReplace: 'a as arg) isFSharp6Supported =
         let lambda = lambda.IgnoreParentParens()
         let binaryExpr = BinaryAppExprNavigator.GetByRightArgument(lambda)
         let argExpr = if isNull binaryExpr then lambda else binaryExpr :> _
         let appTuple = TupleExprNavigator.GetByExpression(argExpr)
         let app = getArgsOwner argExpr
 
         let reference = getReference app
-        app :? IPrefixAppExpr && isNotNull reference &&
+        if not (app :? IPrefixAppExpr) || isNull reference then ctor arg else
```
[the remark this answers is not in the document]
auduchinok: Both examples have an error in from the beginning, we're not really interested in them.

ReSharper.FSharp/src/FSharp.Psi.Daemon/src/Analyzers/LambdaAnalyzer.fs
```diff
@@ -84,45 +84,73 @@ type LambdaAnalyzer() =
 
         compareArgsRec expr 0 null
 
-    let isImplicitlyConvertedToDelegate (lambda: ILambdaExpr) =
+    let getWarningIfApplicable (ctor: ILambdaExpr * 'a -> #IHighlighting) (lambda: ILambdaExpr, exprToReplace: 'a as arg) isFSharp6Supported =
         let lambda = lambda.IgnoreParentParens()
         let binaryExpr = BinaryAppExprNavigator.GetByRightArgument(lambda)
         let argExpr = if isNull binaryExpr then lambda else binaryExpr :> _
```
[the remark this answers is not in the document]
DedSec256: yes, it should work now

ReSharper.FSharp/src/FSharp.Psi.Daemon/src/Analyzers/LambdaAnalyzer.fs
```diff
@@ -84,45 +85,83 @@ type LambdaAnalyzer() =
...
 
-    let isImplicitlyConvertedToDelegate (lambda: ILambdaExpr) =
+    let canBeNamedArg (expr: IBinaryAppExpr) =
+        if isNull expr.Operator then false else
+
+        let exprReference = expr.Operator.Reference
+        if exprReference.GetName() <> "=" then false else
+
+        let declaredElement = exprReference.Resolve().DeclaredElement.As<IFunction>()
+        isNull declaredElement
```
auduchinok, line 95: IIRC `=` is never being resolved when expression is a named arg from the syntax point of view. Here're some examples to consider:

```fsharp
type T =
    static member M(a) = a + 1

T.M(a = 1)

let (=) a b = 1
T.M(a = 1)

let b = 1
T.M(b = 1)
```

Screenshot 2022-08-19 at 18 39 00

ReSharper.FSharp/src/FSharp.Psi.Daemon/src/Analyzers/LambdaAnalyzer.fs
```diff
@@ -84,45 +85,83 @@ type LambdaAnalyzer() =
 
         compareArgsRec expr 0 null
 
-    let isImplicitlyConvertedToDelegate (lambda: ILambdaExpr) =
+    let canBeNamedArg (expr: IBinaryAppExpr) =
+        if isNull expr.Operator then false else
+
+        let exprReference = expr.Operator.Reference
+        if exprReference.GetName() <> "=" then false else
```
auduchinok, line 92: I wonder if we could add `ShortName` to `IBinaryAppExpr`, similar to how it's added to `IReferenceExpr`? It'll hide all the `null` checks.

ReSharper.FSharp/src/FSharp.Psi.Daemon/src/Analyzers/LambdaAnalyzer.fs
```diff
@@ -84,45 +84,73 @@ type LambdaAnalyzer() =
...
-        | :? FSharpMemberOrFunctionOrValue as m ->
-            m.IsMember &&
+        | :? FSharpMemberOrFunctionOrValue as m when m.IsMember ->
             let lambdaPos = if isNotNull appTuple then appTuple.Expressions.IndexOf(argExpr) else 0
 
             let parameterGroups = m.CurriedParameterGroups
-            if parameterGroups.Count = 0 then false else
+            if parameterGroups.Count = 0 then ctor arg else
 
             let args = parameterGroups[0]
```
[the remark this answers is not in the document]
auduchinok: I just realized `param` would also work well, but now it's clear enough anyway 🙂

ReSharper.FSharp/src/FSharp.Psi.Daemon/src/Analyzers/LambdaAnalyzer.fs
DedSec256: 🤔

ReSharper.FSharp/src/FSharp.Psi.Daemon/src/Analyzers/LambdaAnalyzer.fs
DedSec256: i think we can do it in a separate PR

ReSharper.FSharp/src/FSharp.Psi.Features/src/ParameterInfo/FSharpParameterInfo.fs
```diff
@@ -733,27 +734,14 @@ type FSharpParameterInfoContextFactory() =
...
-            match referenceOwner with
-            | :? IFSharpQualifiableReferenceOwner as referenceOwner -> List.ofSeq referenceOwner.Names
-            | _ -> [reference.GetName()]
-
-        let identifier = referenceOwner.FSharpIdentifier
-        if isNull identifier then None else
-
         let endCoords = endOffset.ToDocumentCoords()
         let line = int endCoords.Line + 1
         let column = int endCoords.Column + 1
```
auduchinok, line 739: Coordinates calculation is still duplicated here.
Will `reference.GetElement().GetDocumentEndOffset().ToDocumentCoords()` work?

DedSec256: There is no duplicate here because LambdaAnalyzer calculates coordinates differently

image

auduchinok: So, the only difference is `+ 1` in column expr? That could be a bool param instead of the duplication.
Could you check if not adding `1` could also work for Parameter Info? Or check adding it here? I wouldn't expected difference here to be honest.

auduchinok: The identifier part is interesting for cases when there're type arguments, though...

DedSec256: > Could you check if not adding 1 could also work for Parameter Info?

There are falling tests

> Or check adding it here?

This will also result in a bug

auduchinok: Thanks for checking!

That's sad. We should really find out why the results are different someday.

ReSharper.FSharp/src/FSharp.Psi.Features/src/ParameterInfo/FSharpParameterInfo.fs
```diff
@@ -733,35 +734,18 @@ type FSharpParameterInfoContextFactory() =
...
-        let endCoords = endOffset.ToDocumentCoords()
-        let line = int endCoords.Line + 1
-        let column = int endCoords.Column + 1
-    
-        let checkResults = results.CheckResults
-        match checkResults.GetMethodsAsSymbols(line, column, "", names) with
-        | Some symbolUses when not symbolUses.IsEmpty -> Some(checkResults, symbol, symbolUses)
-        | _ -> Some(checkResults, symbol, [symbolUse])
+        | Some (checkResults, Some symbolUses) when not symbolUses.IsEmpty -> Some(checkResults, symbol, symbolUses)
+        | Some (checkResults, _) -> Some(checkResults, symbol, [symbolUse])
```
auduchinok, line 740:
```suggestion
        | Some (checkResults, []) -> Some(checkResults, symbol, [symbolUse])
        | Some (checkResults, Some symbolUses) -> Some(checkResults, symbol, symbolUses)
```

auduchinok (approved):
Thanks for fixing it!

Also, the new util API is great to have extracted, we'll likely use it in other places. Thanks for taking care of it.

PR #392, opened 2022-08-12
Psi/AbstractMemberDeclaration: fix typeUsage -> returnTypeInfo

DedSec256:
Since we process it as return type info
https://github.com/JetBrains/resharper-fsharp/blob/215b0b4febea04974f01d28555119fa679ad3adc/ReSharper.FSharp/src/FSharp.Psi.Features/src/Parsing/FSharpTreeBuilderBase.fs#L844-L849

https://github.com/JetBrains/resharper-fsharp/blob/215b0b4febea04974f01d28555119fa679ad3adc/ReSharper.FSharp/test/data/parsing/Type%20member%20-%20Abstract%2002%20-%20Function.fs.gold#L18-L36


auduchinok (approved):
Thanks, nice catch!

PR #397, opened 2022-08-18
Build: migrate to JDK 17

ForNeVeR:
- update the target runtime
- update gradle-jvm-wrapper to 0.11 (uses JDK 17 by default)
- regenerate the wrapper
- manually patch the .bat script to unpack tar properly (required until https://github.com/mfilippov/gradle-jvm-wrapper/pull/31 is released)


auduchinok (approved):
Thank you very much!

PR #412, opened 2022-09-08
TypingAssist: generate xmldoc for primary constructor above the type declaration

DedSec256:


DedSec256:
After compilation, documentation for the type is not generated for the primary constructor, so it was decided not to offer typing assist in such cases.

PR #430, opened 2022-10-17
Typing Assist: fix for parameters in a few parens

DedSec256:


ReSharper.FSharp/test/data/features/service/typingAssist/XmlDoc 11 - Parens 01.fs.gold
```diff
@@ -0,0 +1,7 @@
+﻿// ${CHAR:<}
+module Module
+
+/// <summary>
+/// {caret}
+/// </summary>
+let f ((x, y)) = ()
```
auduchinok, line 7: Is it correct that we don't create the param nodes here?

auduchinok (changes requested):
Could you also add tests for methods, constructors, and match expressions in both functions and methods?

ReSharper.FSharp/test/data/features/service/typingAssist/XmlDoc 11 - Parens 01.fs.gold
DedSec256: Sure.

image

auduchinok (approved):

PR #436, opened 2022-10-28
Added intention: Copy and update record field

aciq:
When using nested records in F# it's annoying to copy and update all of them

This action lets you generate a recursive copy and update for a field

Example is here, made an unit test too

```
let newItem = outerRec.Inner.Name[CARET HERE]
```
```
let newItem = { outerRec with Inner = { outerRec.Inner with Name = [CARET HERE] } }
```

IsAvailable checks whether the field belongs to a record


auduchinok (commented):
@aciq Hi, it's nice to see you found a way to fix the annoyance here 🙂
It's an interesting issue indeed!

One thing that I'm concerned about with this approach is its discoverability. What we really want here is to make it easy to find this action, so more people use it.

If you want to have an expression like this:
```fsharp
{ r with Inner = { r.Inner with Field = 1 }
```
Then it's too hard to guess that you need to first write
```fsharp
r.Inner.Field
```
to have this action available.

I was thinking about it for some time, and found another solution that seems interesting. What we can do here is to add a new [postfix template](https://www.jetbrains.com/help/resharper/Postfix_Templates.html), `with`, that would be available on record expressions. That would make it much easier to discover the feature and it would integrate into flow of writing code better.

```fsharp
r.Inner.with
```
can transform code into either
```fsharp
{ r.Inner with {caret} }
```
or
```fsharp
{ r with Inner = { r.Inner with {caret} }
```
It's possible to add a nice popup asking which of these should be used.

This approach is more difficult, on the other hand, but I'd much prefer it if you/we/me went this direction instead. What do you think?

aciq:
@auduchinok A postfix template would be much better, i agree.

It's just that it's currently missing the rest of the infrastructure which i don't know how to add.

Having postfix templates in F# would resolve the issue

auduchinok:
> It's just that it's currently missing the rest of the infrastructure which i don't know how to add.
> Having postfix templates in F# would resolve the issue

The most of the infrastructure is ready and has been there for a while 🙂
It's just we only have `let` template right enabled now, as some of the others indeed need additional work (including in the infra), but I think it should be possible to add the `with` keyword here. [`for`](https://github.com/JetBrains/resharper-fsharp/blob/net223/ReSharper.FSharp/src/FSharp.Psi.Features/src/PostfixTemplates/ForPostfixTemplate.fs) template is a good example to look at. Would you like to give it a try?

aciq:
@auduchinok  Ohh there is an implemented postfix template :o

Sure, i will look into it.

I will report back if i get stuck somewhere

aciq:
@auduchinok i got it working!

now it works with class fields that are records as well

e.g. it turns this
image

to this
image

where the types are like this:

image


i decided to just not allow functions in the template, as it would make multiple calls of the function which i think won't be very useful

so if it finds a function reference in the qualifiers it won't suggest the template

image

i havent made any tests yet and idk how to switch the pull request into the net231 branch, but what do you think?

ReSharper.FSharp/src/FSharp.Psi.Features/src/PostfixTemplates/CopyAndUpdatePostfixTemplate.fs
```diff
@@ -0,0 +1,197 @@
...
+        match recordExpr.FieldBindingList with
+        | null -> recordExpr
+        | bindingList ->
+            let innerChildren =
+                // there is only one child in bindingList
+                bindingList
+                    .Children()
+                    .First()
+                    .As<IRecordFieldBinding>()
+                    .Children()
```
auduchinok, line 105: Please use existing accessors instead of walking the tree manually, e.g. `IRecordLikeExpr.FieldBindings` and `IRecordFieldBinding.Expression`.

auduchinok (commented):
@aciq Very nice change, and it's really cool that you've managed to make it work!

> i decided to just not allow functions in the template, as it would make multiple calls of the function which i think won't be very useful

To be honest, I'd expect it to work: from as user standpoint it's not obvious why the template works in some cases and doesn't in others. Especially, since it's working for other kinds of invocations like indexers, computable properties and so on. To make it work we should be dealing with expression tree nodes themselves and not with identifiers texts in `createCopyAndUpdate`, i.e. get qualifier expression nodes, move them, and clone where needed.

To make this work in more user cases I suggest trying to add a popup asking how many levels a user wants to update: it seems it may be useful to update the last field only (and get a new value that is not related to the old value tree) or some of the fields (but it would be great to test it in real life a bit), and then use the new value in another record value or a binding. Could you try using a logic similar to what's done in `AddIgnoreFix` and Introduce variable refactoring?

Here're some examples:
https://github.com/JetBrains/resharper-fsharp/blob/614b38aa4ab83f7bc37cdfcb2993149565cf225e/ReSharper.FSharp/src/FSharp.Psi.Intentions/src/QuickFixes/AddIgnoreFix.fs#L60
https://github.com/JetBrains/resharper-fsharp/blob/614b38aa4ab83f7bc37cdfcb2993149565cf225e/ReSharper.FSharp/src/FSharp.Psi.Intentions/src/QuickFixes/AddParensToApplicationFix.fs#L124

For your example it'd suggest these variants:
```
Record.R2.R1.Name
R2.R1.Name
R1.Name
Name
```

What do you think?

aciq:
> To be honest, I'd expect it to work: from as user standpoint it's not obvious why the template works in some cases and doesn't in others. Especially, since it's working for other kinds of invocations like indexers, computable properties and so on. To make it work we should be dealing with expression tree nodes themselves and not with identifiers texts in createCopyAndUpdate, i.e. get qualifier expression nodes, move them, and clone where needed.

I thought about it and i think you're right on that actually, it's probably best if it works in all cases

> To make this work in more user cases I suggest trying to add a popup asking how many levels a user wants to update: it seems it may be useful to update the last field only (and get a new value that is not related to the old value tree) or some of the fields (but it would be great to test it in real life a bit), and then use the new value in another record value or a binding. Could you try using a logic similar to what's done in AddIgnoreFix and Introduce variable refactoring?

In this case is it possible to have a keybind for them? i'm thinking the two most common uses for me would be updating the entire record tree or updating only the last record so it would be very nice to have a keyboard shortcut without the popup for these. 

I havent yet figured out how creating RecordExprs works, i know i could create one with tokens individually but is there a utility function for this?

aciq:
Ok i like this a lot already, here's how it works now.

copy_and_update_1

One thing that i didnt figure out is how i can create a IRecordFieldBindingList, so i currently just passed in a single IRecordFieldBinding directly

aciq:
i have one more question, how do i turn this feature on in the IDE?

i only got it working by forcing this to be true
```fsharp
    override this.IsEnabled(solution) = 
        let configurations =
            solution.GetComponent<RunsProducts.ProductConfigurations>()
        
        configurations.IsInternalMode()
        || ``base``.IsEnabled(solution)

```

auduchinok:
> Ok i like this a lot already, here's how it works now.

Looks really cool, indeed!

> i have one more question, how do i turn this feature on in the IDE?

Just return `true` in `IsEnabled`. This method was added to allow enabling unfinished templates for testing.
https://github.com/JetBrains/resharper-fsharp/blob/9f864d97fd7c72bee130a3908e936d8587047e4b/ReSharper.FSharp/src/FSharp.Psi.Features/src/PostfixTemplates/LetPostfixTemplate.fs#L24

> One thing that i didnt figure out is how i can create a IRecordFieldBindingList, so i currently just passed in a single IRecordFieldBinding directly

You can also get it from the sample record expression, `IRecordLikeExpr.FieldBindingList`.
A bit off-topic: I'm going to remove this intermediate node at some point, it was added for an old formatter prototype, and we don't need it anymore.

auduchinok:
The more I look at it, the more I think we could also try to make it a bit different: enable it on record values instead of record fields.

So this example:
```fsharp
r3.R2.R1.with{caret}
```
will expand into this
```fsharp
{ r3.R2.R1 with {caret} }
```

It seems a bit more natural if you think about 1) `with` being actually applied to a record, not to a field, and 2) about updating several fields, like:

```fsharp
{ getProps() with
    Field1 = true
    Field2 = "" }
```

And then we could automatically show the completion popup, so all relevant fields are immediately visible (2022.3 already moves them to the top of the popup). To make it even better we would need some experimenting with code completion, like appending ` = ` to a field automatically.

If you don't mind experimenting with it, you can use the following check:
```fsharp
let isRecord (expr: IFSharpExpression) =
    let fcsType = expr.TryGetFcsType()
    if isNull fcsType then false else

    let fcsType = getAbbreviatedType fcsType
    if not fcsType.HasTypeDefinition then false else

    fcsType.TypeDefinition.IsFSharpRecord
```

Getting field qualifiers would still allow updating nested records, so it should be good.

If we decide to make it work this way, I can do the field completion part later if we find some difficulties now.

aciq:
yeah it's a good idea, plus the "with" expression constrains the available fields in autocomplete as well, so i see no drawbacks.

one thing that usually bothers me about unspecified record expressions is that you get suggestions for every possible record.

i will play around with it for a bit

aciq:
ok so this works great, i noticed some more things.

recordvalue

> And then we could automatically show the completion popup, so all relevant fields are immediately visible (2022.3 already moves them to the top of the popup). To make it even better we would need some experimenting with code completion, like appending = to a field automatically when completion item is accepted.

first this would definitely make it even better, i like both the idea of immediate transition into completions and the `=`.

and the second thing is that i should constrain the available items if it already knows the resulting type - so in the following example i already know the resulting type is R1B, so this should be the only available option. and if there's only one option it won't even show the menu, so it's all very smooth. 

knowntype

though that might kill the option for mapping other records into the R1B type. so perhaps i should constrain it only if one of the options is an exact match of the known result type. i'll test it for a bit longer and see

aciq:
also perhaps it would be clearer if the options were just
```
r3
R2
R1B
```
instead

aciq:
this isnt exactly related to this task but how would you replace the entire raw text contents of a file? i want to rewrite the whole syntax tree with a string and then walk over the nodes once it's done.

i found that you can get the file root like this but i got stuck there
```
let fsfile = textControl.GetFSharpFile(sln)
let root = fsfile.Root()
```

auduchinok:
> and the second thing is that i should constrain the available items if it already knows the resulting type - so in the following example i already know the resulting type is R1B, so this should be the only available option. and if there's only one option it won't even show the menu, so it's all very smooth.

I'd prefer not filtering the variants, at least for the start, as it would:
* allow using a different type that would influence type inference
* allow using a wrong type and changing the field declaration with a quick fix later
* be a bit more consistent

> also perhaps it would be clearer if the options were just instead

Let's try to highlight them in the editor instead? It's possible to pass a highlighting range in `WorkflowPopupMenuOccurrence` constructor.


@aciq Could you please write some tests, and I'll do another review after that?

auduchinok:
> this isnt exactly related to this task but how would you replace the entire raw text contents of a file? i want to rewrite the whole syntax tree with a string and then walk over the nodes once it's done.

This is not what we normally do in quick fixes, context actions, and other features where we work with tree nodes instead. There're various reasons to do it, including these:
* R# can do a lot of additional changes that fix the formatting and code style along the way (we don't use it in F# yet, but we're going to)
* there're various caches stored in nodes themselves that are kept until a node changes, and replacing the whole file would invalidate extra things (we also lose some of these optimizations in F# already because of the lack of incremental reparse and resolve isolation scopes, but it's better to not add more places that prevent optimizations)

One way to achieve this is to get a new tree with an element factory and then replace the root node. Another option is to replace the document text, commit changes and get a new tree, see `FSharpReformatCode` for an example.

Could you describe what you want to achieve here? Feel free to ping @auduchinok on F# Slack or Telegram if you'd like to chat about it.

aciq:
@auduchinok sure, i would look at the tests next week when i have a windows machine again if that's okay.

aasiki:
@auduchinok i think i messed up this pull request's history by merging from net223. i think i'll make a new pull request starting from the net231 branch once i've got working tests

auduchinok (commented):
@varatalu I've fixed the logic for getting copy expression, added completion reschedule, and added tests. I think it looks good to get in now:

postfixWith

Thanks again for bringing this topic and proposing ideas how to deal with it.

varatalu:
@auduchinok okay, very good!

PR #449, opened 2022-11-23
ExprStaticConstantTypeUsage: add navigator for constant expression type arg

DedSec256:
At https://github.com/JetBrains/resharper-fsharp/pull/134 we should be able to get ExprStaticConstantTypeUsage by its literal parameter, which is not a chameleon 
https://github.com/JetBrains/resharper-fsharp/blob/a8e7fb2237718f64001274029edb9b45721b7732/ReSharper.FSharp/src/FSharp.Psi.Features/src/Parsing/FSharpTreeBuilderBase.fs#L804-L810


ReSharper.FSharp/src/FSharp.Psi/src/FSharp.psi
```diff
@@ -1807,7 +1807,11 @@ exprStaticConstantTypeUsage extras {
   get { methodName="Expression" path=<exprStaticConstantTypeUsage:CHAMELEON_EXPR/chameleonExpression:EXPR> };
 }:
   CONST?
-  chameleonExpression<CHAMELEON_EXPR, ChameleonExpression>;
+  (
+    chameleonExpression<CHAMELEON_EXPR, ChameleonExpression>
```
auduchinok, line 1811: In what cases there's a chameleon and in what cases there's an expression?

DedSec256: `A<"123">`
image

`A<const("123")>`
image

auduchinok: Why do we have different trees for these expressions? Looks like a tree builder bug to me

DedSec256: Perhaps there were differences, because in the case of a literal in the compiler tree, only range is returned, not expr. 
It can be unified, but it seemed redundant to me to create a chameleon for a simple literal.

auduchinok: I agree it seems redundant.

What I'm concerned about is in all other places we wrap expressions inside chameleons, even simple ones. We could stop doing it in some cases like this, but right now this case is just too inconsistent with the rest of the tree.

It helps when you have some simple/consistent rules like 'all top level expressions are wrapped in chameleons', but it turns out it's not the case, and now everyone has to take into account very rare cases where expressions aren't wrapped.

ReSharper.FSharp/src/FSharp.Psi.Features/src/Parsing/FSharpTreeBuilderBase.fs
```diff
@@ -806,8 +806,10 @@ type FSharpTreeBuilderBase(lexer, document: IDocument, lifetime, path: VirtualFi
             x.MarkChameleonExpression(expr)
             x.Done(range, mark, ElementType.EXPR_STATIC_CONSTANT_TYPE_USAGE)
 
-        | SynType.StaticConstant(synConst, _) ->
-            x.MarkAndDone(range, ElementType.EXPR_STATIC_CONSTANT_TYPE_USAGE, x.GetConstElementType(synConst))
+        | SynType.StaticConstant(synConst, constRange) ->
+            let mark = x.Mark(range)
+            x.MarkChameleonExpression(SynExpr.Const(synConst, constRange))
```
auduchinok, line 811: Clever trick!

auduchinok (approved):
Thanks!

PR #463, opened 2022-12-17
Remove unused record named pattern quick fix

edgarfgp:
This PR adds a new quick to `Remove unused record named pattern quick fix` . This can be extended to work with named union types in a separate PR :) 

RemoveUnusedRecordNamedPattern


ReSharper.FSharp/src/FSharp.Psi.Intentions/src/QuickFixes/RemoveUnusedNamedPatternFix.fs
```diff
@@ -0,0 +1,75 @@
...
+
+type RemoveUnusedNamedPatternFix(warning: UnusedValueWarning) =
+    inherit FSharpQuickFixBase()
+    
+    let fieldAndRecord =
+        match warning.Pat.IgnoreInnerParens().Parent with
+        | :? IFieldPat as fieldPat ->
+            match fieldPat.Parent with
+            | :? IRecordPat as recordPat ->
+                Some(fieldPat, recordPat)
```
auduchinok, line 22: Please use generated navigation helpers instead of manually traversing the tree: it's harder to make a mistake and it'll make changing the tree structure much easier later, and we often do it for some parts of the tree. `RecordPatNavigator.GetByFieldPattern` is what you'd want to use here.

ReSharper.FSharp/test/data/features/quickFixes/removeUnusedNamedPattern/Unused record named pattern first position.fs
```diff
@@ -0,0 +1,5 @@
+type R = { F1: int; F2: int; F3: int }
+match { F1 = 1; F2 = 2; F3 = 3 } with
+| { F1 = f1{caret}; F2 = 2; F3 = 3 } -> ()
+| { F1 = 1; F2 = f2; F3 = 3 } -> ()
+| { F1 = f1; F2 = f2; F3 = f3 } -> ()
```
auduchinok, line 5: Please remove the extra match branches: they don't add anything useful to the test while make it harder to see the important details from a quick glance.
```suggestion
type R = { F1: int; F2: int }
match { F1 = 1; F2 = 2 } with
| { F1 = f1{caret}; F2 = 2 } -> ()
```

ReSharper.FSharp/src/FSharp.Psi.Intentions/src/QuickFixes/RemoveUnusedNamedPatternFix.fs
```diff
@@ -0,0 +1,75 @@
...
+            null
+            
+        elif indexPat < recordPat.FieldPatterns.Count - 1 then
+            let nextFieldPat = recordPat.FieldPatterns.[indexPat + 1]
+            
+            let rangeToDelete =
+                let rangeStart = fieldPat
+                let rangeEnd =
+                    nextFieldPat.PrevSibling
+                TreeRange(rangeStart, rangeEnd)
```
auduchinok, line 57: This approach doesn't look at what could be between the field patterns, and there are some important cases to consider when using it:

We probably don't want to remove the new line between the fields here:
```fsharp
match r with
| { F1 = 1; F2 = f2
    F3 = 3; F4 = 4 } -> ()
```

While here we would not want to keep an empty line:
```fsharp
match r with
| { F1 = 1
    F2 = f2
    F3 = 3 } -> ()
```

What I'd suggest trying to do here is this:
* get the field pattern node
* check if it's the last field pattern on the line, i.e. if `fieldPat.EndLine <> nextPattern.StartLine`
  * if it's the last one: try to go left from it and possibly include the previous new line token, like in the last example, using `getFirstMatchingNodeBefore isInlineSpace fieldPat |> getThisOrPrevNewLine` helpers
* try to move to the right using `getLastMatchingNodeAfter isInlineSpace`

Please check existing usages of these helpers.

ReSharper.FSharp/src/FSharp.Psi.Intentions/src/QuickFixes/RemoveUnusedNamedPatternFix.fs
```diff
@@ -0,0 +1,75 @@
...
+            
+            ModificationUtil.DeleteChildRange(rangeToDelete)
+            
+            let offset =nextFieldPat.GetNavigationRange().StartOffsetRange().StartOffset
+            Action<_>(fun textControl ->
+                textControl.Caret.MoveTo(offset, CaretVisualPlacement.DontScrollIfVisible))
```
auduchinok, line 63: Maybe this isn't required, will it place the caret itself good enough if you just return null?

ReSharper.FSharp/src/FSharp.Psi.Intentions/src/QuickFixes/RemoveUnusedNamedPatternFix.fs
auduchinok: Some more cases to add a test for:

```fsharp
match r with
| { F1 = 1; F2 = 2
    F3 = f3; F4 = 4 } -> ()
```

```fsharp
match r with
| { F1 = 1; F2 = 2
    F3 = f3 } -> ()
```

```fsharp
match r with
| { F1 = 1; F2 = f2 // comment
    F3 = 3 } -> ()
```

auduchinok (commented):
@edgarfgp This is nice work!

The quick fix may remove extra nodes in some cases, and it would be better not to do it. I left a suggestion on how to better calculate the tree range to remove.

ReSharper.FSharp/src/FSharp.Psi.Services/src/Daemon/Highlightings/FcsErrors.xml
```diff
@@ -92,6 +92,7 @@
     <QuickFix>ReplaceWithWildPatFix</QuickFix>
     <QuickFix>RemoveUnusedNamedAsPatFix</QuickFix>
     <QuickFix>AddUnderscorePrefixFix</QuickFix>
+    <QuickFix>RemoveUnusedNamedPatternFix</QuickFix>
```
auduchinok, line 95: The quick fix is currently at the end of the list but is arguably much more useful than `ReplaceWithWildPatFix` and `AddUnderscorePrefixFix `. Let's move it up? 🙂

ReSharper.FSharp/src/FSharp.Psi.Intentions/src/QuickFixes/RemoveUnusedNamedPatternFix.fs
```diff
@@ -0,0 +1,75 @@
...
+open JetBrains.ReSharper.Psi.Util
+open JetBrains.ReSharper.Resources.Shell
+open JetBrains.TextControl
+
+type RemoveUnusedNamedPatternFix(warning: UnusedValueWarning) =
+    inherit FSharpQuickFixBase()
+    
+    let fieldAndRecord =
+        match warning.Pat.IgnoreInnerParens().Parent with
+        | :? IFieldPat as fieldPat ->
```
auduchinok, line 19: Please also use `FieldPatNavigator.GetByPattern` here.

ReSharper.FSharp/src/FSharp.Psi.Intentions/src/QuickFixes/RemoveUnusedNamedPatternFix.fs
edgarfgp: Yes it looks like it is enough

ReSharper.FSharp/test/data/features/quickFixes/removeUnusedNamedPattern/Unused record named pattern with a single field.fs.gold
```diff
@@ -1,3 +1,3 @@
 type R = { F1: int }
 match { F1 = 1 } with
-| _{caret} -> ()
+| { F1 = _{caret} } -> ()
```
auduchinok, line 3: This change looks unexpected: there's another quick fix that does it, and it's safe to remove the record pattern in this case.

auduchinok: When removing the last record field pattern, like in this test.

ReSharper.FSharp/test/data/features/quickFixes/removeUnusedNamedPattern/Unused union case named pattern 1.fs.gold
```diff
@@ -0,0 +1,3 @@
+type R = | R of a: int
+match R(0) with
+| _{caret} -> ()
```
auduchinok, line 3: This changes the code meaning if when there're other branches or union cases. This is the case we've discussed previously. You should not remove the whole union case pattern in this case, only its fields.

The value of `x` is going to be different here after this quick fix is applied:
```fsharp
type U =
    | A of i: int
    | B

let x =
    match A(1) with
    | A(i = i) -> 1
    | B -> 2
```

auduchinok (changes requested):

ReSharper.FSharp/test/data/features/quickFixes/removeUnusedNamedPattern/Unused union case named pattern 1.fs.gold
edgarfgp: Ok I see so in this case this case after applying the quick fix the result would be : 
```fsharp
let x =
    match B with
    | A  _ -> 1
    | B -> 2
```

auduchinok (changes requested):

ReSharper.FSharp/test/data/features/quickFixes/removeUnusedNamedPattern/Unused union case named pattern 1.fs.gold
auduchinok: Yes.

PR #468, opened 2023-01-02
LambdaAnalyzer: disable for invocations with optional args

DedSec256:
Fix for [RIDER-87136](https://youtrack.jetbrains.com/issue/RIDER-87136/False-positive-simplify-lambda)


auduchinok (approved):
Thanks!

PR #474, opened 2023-01-12
Fixed highlighting for xint literals with separators

En3Tho:
Changes lexer a little bit to properly highlight numbers with prefix and separators like
```fsharp
0b1_1 0o1_1 0x1_1
```
Before: highlighted as text
After: highlighted as number


En3Tho:
@auduchinok Sure. For FSharpLexerTest - do you mean that instead of adding those to testDigit I should create another test specifically for underscores or should I just multiple underscore cases in addition to single ones?

auduchinok (approved):
@En3Tho Thanks, this is a very good PR!

PR #475, opened 2023-01-24
Frontend/Tests/Type Providers: fix changeStaticArg test

DedSec256:


auduchinok (approved):
Waiting for SWEA is probably an overkill here, but if it makes the test more stable I'm OK with merging it as is at least for now

PR #476, opened 2023-01-24
Reformat Kotlin code

ForNeVeR:
This brings the code format closer to what we have in the internal **dotnet-products** monorepo, and what we have in the [IntelliJ Community](https://github.com/JetBrains/intellij-community). So, it will be easier to exchange code snippets if the need arises.

Besides, I've added `.git-blame-ignore-revs` that's supposed to exclude this huge re-formatting commit from `git blame` and similar operations. [Read about it here](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view), see [an example of how the blame looks on GitHub here](https://github.com/JetBrains/resharper-fsharp/blame/d39c677d779c84ea9ee03996302db0198c6bbd60/rider-fsharp/src/main/java/com/jetbrains/rider/plugins/fsharp/breadcrumbs/breadcrumbs.kt).

**Note** that additional Git setup will be required for this file to be taken into account locally (it only gets auto-used by GitHub and some similar tools); see the instructions in the file itself. Sadly, it's impossible to share the Git settings together with the repository.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,7 @@
+; Topmost EditorConfig file
+root = true
+
+[*.{java,kt,kts}]
+indent_style = space
+indent_size = 2
+tab_width = 2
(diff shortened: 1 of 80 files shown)


auduchinok (approved):
@ForNeVeR Thank you a lot for this!

PR #478, opened 2023-02-09
Localize Errors.xml and FcsErrors.xml

retran:
This PR:
- adds a new resources file with strings from Errors.xml and FcsErrors.xm;
- converts Errors.xml and FcsErrors.xml to a new format that uses strings from resources.


auduchinok (approved):
@retran Thank you for implementing the generation and updating these files! 🙂

PR #480, opened 2023-02-10
XmlDocs: disable highlightings in simple summary doc comments

DedSec256:
Since they are interpreted as plaintext


DedSec256:
I think we can do something like this

image

PR #481, opened 2023-02-13
Initial generation signature feature toggle.

nojaf:
A second attempt at https://github.com/JetBrains/resharper-fsharp/pull/378.


ReSharper.FSharp/src/FSharp.Psi.Intentions/src/Intentions/GenerateSignatureFileAction.fs
```diff
@@ -0,0 +1,100 @@
...
+        let rec processModuleMembers (parent: IFSharpTreeNode) (members: IModuleMember seq) =
+            for m in members do
+                match m with
+                | :? INestedModuleDeclaration as nmd ->
+                    let nestedModuleNode = factory.CreateNestedModule(nmd.NameIdentifier.Name)
+                    // TODO: if the nested module has nested modules, clear the content (`begin end`) and process them.
+                    ModificationUtil.AddChild(parent, nestedModuleNode) |> ignore
+                | _ -> ()
+        
+        for decl in fsharpFile.ModuleDeclarations do
```
nojaf, line 44: @auduchinok I know you mentioned using the `TreeNodeVisitor` here, but I don't quite see how to easily use that here. Any suggestions?

auduchinok: It's OK to do it this way too, maybe it'll be even easier at the end

ReSharper.FSharp/test/data/features/generate/signatureFiles/Module structure 02.fs
```diff
@@ -0,0 +1,11 @@
+// ${KIND:SignatureFile}
+// ${SELECT0:Generate signature file title}
+module Foo
+
+open System
+    let c = Math.PI
```
nojaf, line 6: Bit of a nitpick but this indent seems unnecessary.

ReSharper.FSharp/src/FSharp.Psi.Services/src/Generate/GenerateSignatureProvider.fs
```diff
@@ -146,6 +146,29 @@ type FSharpGenerateSignatureBuilder() =
...
+                        let symbolUse = refPat.GetFcsSymbolUse()
+                        if isNotNull symbolUse then
+                            let mfv = symbolUse.Symbol :?> FSharpMemberOrFunctionOrValue
+                            sb.Append(mfv.FullType.Format(symbolUse.DisplayContext)) |> ignore
+                  
+                    sb.ToString()
+
+                let sigStrings =
+                    Seq.map sourceString letBindingsDeclaration.Bindings
+                    |> String.concat System.Environment.NewLine
```
nojaf, line 169: I think we can just use `"\n"` in this case, as Rider internally uses `lf` as the line ending.
Does this matter @auduchinok?

ReSharper.FSharp/src/FSharp.Psi.Services/src/Generate/GenerateSignatureProvider.fs
```diff
@@ -0,0 +1,312 @@
...
+        signatureFile
+
+    override this.IsAvailable(context: FSharpGeneratorContext): bool =
+        let node = context.Root :?> IFSharpTreeNode
+        let currentFSharpFile = node.FSharpFile
+        if currentFSharpFile.IsFSharpSigFile() then false else
+
+        let solution = node.GetSolution()
+        let isSettingEnabled = solution.IsFSharpExperimentalFeatureEnabled(ExperimentalFeature.GenerateSignatureFile)
+        if not isSettingEnabled then false else
```
auduchinok, line 256: It seems this place is being checked a bit too late: the workflow gets shown inside Alt+Enter menu and is just getting disabled. We could try to use the `dataContext` that is passed to `FSharpGenerateSignatureWorkflowProvider.CreateWorkflow`, get a solution from it and check if the setting is enabled.

ReSharper.FSharp/src/FSharp.Psi.Services/src/Generate/GenerateSignatureProvider.fs
auduchinok: I think it doesn't matter much, at least for now. I don't know the preferred way when creating new files, but I'd start with reusing new lines from the existing file, i.e. just use `lineEnding` from line 62.

We may also start supporting this plugin in ReSharper in VS future, and I don't think that they have this "always `"\n"`" logic there.

ReSharper.FSharp/src/FSharp.Psi.Services/src/Generate/GenerateSignatureProvider.fs
```diff
@@ -192,7 +192,12 @@ type FSharpGenerateSignatureBuilder() =
                 factory.CreateTypeMemberSignature(sigStrings)
             | :? IExceptionDeclaration as exceptionDeclaration ->
                 let sigExceptionDeclaration = exceptionDeclaration.Copy()
-                if not (Seq.isEmpty exceptionDeclaration.MemberDeclarations) then
+                let memberDeclarations =
+                    exceptionDeclaration.MemberDeclarations
+                    |> Seq.filter (function | :? IExceptionFieldDeclaration -> false | _ -> true)
+                    |> Seq.toArray
```
auduchinok, line 198: `exceptionDeclaration.TypeMember` is probably what you're after here.

ReSharper.FSharp/src/FSharp.Psi.Services/src/Generate/GenerateSignatureProvider.fs
```diff
@@ -266,15 +266,29 @@ type FSharpGenerateSignatureBuilder() =
...
-                | :? IUnitPat ->
-                    $"new: unit -> {typeName}"
-                | _ -> failwithf "todo"
+        and createPrimaryConstructorSignature (typeName: string) (primaryConstructorDeclaration: IPrimaryConstructorDeclaration) : ITreeNode seq =
+            match primaryConstructorDeclaration.ParameterPatterns with
+            | :? IUnitPat ->
+                factory.CreateTypeMemberSignature $"new: unit -> {typeName}"
+                :> ITreeNode
+                |> Seq.singleton 
+            | _ ->
```
auduchinok, line 280: Is this check actually required? I'd expect getting `FcsSymbolUse` would work for empty primary constructors too.

ReSharper.FSharp/src/FSharp.Psi.Services/src/Generate/GenerateSignatureProvider.fs
nojaf: Good catch, I think I solved it.

ReSharper.FSharp/src/FSharp.Psi.Services/src/Generate/GenerateSignatureProvider.fs
nojaf: Yup

ReSharper.FSharp/src/FSharp.Psi.Services/src/Generate/GenerateSignatureProvider.fs
```diff
@@ -171,15 +171,18 @@ type FSharpGenerateSignatureBuilder() =
...
-                    sb.Append(": ") |> ignore
-                    
                     let refPat = binding.HeadPattern.As<IReferencePat>()
                     if isNotNull refPat then
                         let symbolUse = refPat.GetFcsSymbolUse()
                         if isNotNull symbolUse then
                             let mfv = symbolUse.Symbol :?> FSharpMemberOrFunctionOrValue
+
+                            mfv.Attributes
+                            |> Seq.iter (fun a -> sb.Append($"{a.Format(symbolUse.DisplayContext)}{lineEnding}") |> ignore)
```
nojaf, line 181: What happens when the attribute has arguments?
`[<Ignore "Foobar">]`, I'd be surprised if the Format takes care of that.
Does the factory generated member have a `SetAttribute` call instead?

dawedawe: I adjusted the test. I think Format is doing what we need regarding arguments.

nojaf: Until it won't, try
```fsharp
module Telplin

type MyEnum =
    | A = 0
    | B = 1

type BAttribute(v:MyEnum) =
    inherit System.Attribute()

[<BAttribute(MyEnum.A)>]
let a b c = b + c
```

I think this won't work out.
The safer option would be to reuse the attribute from the implementation file.

auduchinok: How does it work with named literal values?

```fsharp
[<Flags>]
type Enum =
    | None = 0
    | Aa = 1
    | Bb = 2
    | Cc = 4
    | AaAndCc = 5

type AttrAttribute() =
    inherit Attribute()

    member val Prop = Enum.None with get, set 

[<Attr(Prop = Enum.AaAndCc)>]
let a = 1
```

```fsharp
let [<Literal>] Hello = "Hello"

[<CompiledName(Hello)>]
let hello = 1
```

nojaf: Yup, https://github.com/dotnet/fsharp/issues/13810

auduchinok: > Yup, https://github.com/dotnet/fsharp/issues/13810

IIRC FCS doesn't preserve the reference to a named literal and only captures the literal value. So if you have multiple literals with the same value, it wouldn't know which one to use.

Consider this case:
```
let [<Literal>] Hello1 = "Hello"
let [<Literal>] Hello2 = "Hello"

[<CompiledName(Hello1)>]
let hello = 1
```

We want to preserve the named literal reference, so it's possible to update the attribute values in both signature and implementation files by changing the single named literal value. And it'll also prevent features like Find Usages and Rename from breaking.

> The safer option would be to reuse the attribute from the implementation file.

This looks like the way to go, yes.

dawedawe: Hehe, you guys know your stuff.
So I used `GetText()` to get the original attribute from the implementation file. Is that okay?
I still need to make `IEnumRepresentation` work...

nojaf: This is not what I meant. So, `binding` has a bunch of setters to insert existing nodes into it.
Like
image


And wishful thinking the newly created `IModuleMember` had something similar like `SetAttribute`
@auduchinok any ideas?

ReSharper.FSharp/src/FSharp.Psi.Services/src/Generate/GenerateSignatureProvider.fs
```diff
@@ -177,8 +177,8 @@ type FSharpGenerateSignatureBuilder() =
                         if isNotNull symbolUse then
                             let mfv = symbolUse.Symbol :?> FSharpMemberOrFunctionOrValue
 
-                            mfv.Attributes
-                            |> Seq.iter (fun a -> sb.Append($"{a.Format(symbolUse.DisplayContext)}{lineEnding}") |> ignore)
+                            binding.Attributes
+                            |> Seq.iter (fun a -> sb.Append($"[<{a.GetText()}>]{lineEnding}") |> ignore)
```
auduchinok, line 181: Why do you prefer to do a fresh parsing instead of copying the nodes?
Does it handle attribute targets and indentation properly?

ReSharper.FSharp/src/FSharp.Psi.Services/src/Generate/GenerateSignatureProvider.fs
auduchinok: > And wishful thinking the newly created IModuleMember had something similar like SetAttribute

For some reason the generator didn't create it. I'd suggest to work with tree manually until we can enable the formatter to handle all the whitespace-related modifications for us.

nojaf: By just inserting the nodes before the val node?
image

ReSharper.FSharp/src/FSharp.Psi.Services/src/Generate/GenerateSignatureProvider.fs
dawedawe: Sorry, could you elaborate? I'm pretty much out of my depth here.
How should I copy the nodes to build up the string we need?

auduchinok: It seems there should be no reason to build a string of the attributes in the first place, if you already have the desired nodes. Please see what this PR already does for `open` nodes.

PR #488, opened 2023-03-02
Add .md file with documentation about how to dump the PSI to a file

dawedawe:
First part of #487


docs/dump_psi_to_file.md
```diff
@@ -0,0 +1,9 @@
+## Dumping the PSI to a file
```
auduchinok, line 1: It's probably worth noting that this is about R# trees, not about IntelliJ ones. Maybe we should have separate docs (separate folders?) for these two plugins, and an additional doc telling which plugin does what.

It's not relevant just yet, but once https://github.com/JetBrains/resharper-fsharp/pull/482 is merged, it'd be helpful to have docs about dumping those trees as well. It can be done like this (and it requires [Internal mode](https://rider-support.jetbrains.com/hc/en-us/articles/207327910-How-to-run-JetBrains-Rider-in-Internal-Mode)):

Screenshot 2023-03-02 at 11 32 41

dawedawe: > It's probably worth noting that this is about R# trees, not about IntelliJ ones. Maybe we should have separate docs (separate folders?) for these two plugins, and an additional doc telling which plugin does what.
> 
Okay, I moved the new `docs` folder down to `ReSharper.FSharp`.
About the distinction of the two plugins, I think that's already covered in the existing `README.md`.

> It's not relevant just yet, but once #482 is merged, it'd be helpful to have docs about dumping IntelliJ trees as well. It can be done like this (and it requires [Internal mode](https://rider-support.jetbrains.com/hc/en-us/articles/207327910-How-to-run-JetBrains-Rider-in-Internal-Mode), it's automatically used when running the plugin via Gradle):

Sure, once that's done, I can provide docs for that, too :)

README.md
```diff
@@ -73,6 +73,7 @@ We suggest that you read docs on the two SDKs that this plugin uses:
 * [ReSharper SDK](https://www.jetbrains.com/help/resharper/sdk/welcome.html)
 * [IntelliJ IDEA SDK](https://www.jetbrains.org/intellij/sdk/docs/welcome.html)
 
+You can find some tips and tricks about the ReSharper.Host plugin (the *backend*) in the `ReSharper.FSharp\docs` folder of this repository.
```
auduchinok, line 76: Will a link work correctly?
```suggestion
You can find some tips and tricks about the ReSharper.Host plugin (the *backend*) in the [ReSharper.FSharp/docs](ReSharper.FSharp/docs) folder of this repository.
```

auduchinok (approved):
@dawedawe Thanks!

PR #493, opened 2023-03-05
Use named access intention

nojaf:
I was tinkering around with this intention:

namedAccess

@auduchinok I did find something weird in the element factory for `INamedUnionCaseFieldsPat`, the closing right parenthesis doesn't seem to be a part of it.

image


auduchinok:
> @auduchinok I did find something weird in the element factory for INamedUnionCaseFieldsPat, the closing right parenthesis doesn't seem to be a part of it.

Is the paren included in the needed range in the FCS syntax tree? If not, it should probably be fixed there. If yes, then it's likely some issue in `FSharpImplTreeBuilder`.

ReSharper.FSharp/src/FSharp.Psi.Intentions/src/Intentions/UseNamedAccessAction.fs
```diff
@@ -0,0 +1,107 @@
...
+        
+        let hasUnderscore =
+            tuplePat.Patterns
+            |> Seq.exists (fun p ->
+                match p with
+                | :? IWildPat -> true
+                | _ -> false)
+    
+        // We need something that is currently ignored
+        if not hasUnderscore then false else
```
auduchinok, line 43: That's an interesting choice. 🙂 

Let's consider these cases:
```fsharp
match x with
| UnionType.UnionCase1(true, true, true) -> ()
| UnionType.UnionCase1(fieldName1, fieldName2, _) -> ()
```

In the first case it might make sense to enable the action, as the use of literals make it a bit harder to know what fields they match. But it's not enabled because there's no `_` patterns.

In the second case it's enabled, but I'd consider it much less useful, as the names already may say what they stand for (and pattern generation features would try to use names from the fields if possible).


The range where this action is available is also worth discussing. Normally, we try to limit such ranges because there are more and more actions being added over years, and we don't want later neither 1) to have to many at one place 2) to restrict the ranges and break habits. A union name and `_` patterns could be good ranges instead, but _maybe_ for this particular action it's worth to have a single longer range. It definitely needs some experimenting.

nojaf:
> Is the paren included in the needed range in the FCS syntax tree?

Yes, I believe so.

image

ReSharper.FSharp/src/FSharp.Psi.Intentions/src/Intentions/UseNamedAccessAction.fs
```diff
@@ -0,0 +1,107 @@
...
+        // This should only work when there are multiple items
+        tuplePat <- parenPat.Pattern :?> ITuplePat
+        if isNull tuplePat then false else
+        
+        let hasUnderscore =
+            tuplePat.Patterns
+            |> Seq.exists (fun p ->
+                match p with
+                | :? IWildPat -> true
+                | _ -> false)
```
auduchinok, line 40: Just in case if we want to consider more cases later:

```fsharp
// `_` is nested inside `asPat`
| A(a, (b, c) as _) -> ()

// `_` is nested inside `tuplePat`
| A(a, (_, _)) -> ()
```

ReSharper.FSharp/src/FSharp.Psi.Intentions/src/Intentions/UseNamedAccessAction.fs
```diff
@@ -0,0 +1,107 @@
...
+        // All fields need to be named
+        if names.Length <> fcsUnionCase.Fields.Count then false else
+        // The amount of fields needs to match with the tuple length
+        if names.Length <> tuplePat.Patterns.Count then false else
+
+        true
+
+    override x.ExecutePsiTransaction(_,_) =
+        let pattern = dataProvider.GetSelectedElement<IParametersOwnerPat>()
+        if isNull pattern then null else
```
auduchinok, line 63: Please throw an exception instead. This code should only ever run if `IsAvailable` is `true`, and that would mean the pattern should be accessible.
```suggestion
        let pattern = dataProvider.GetSelectedElement<IParametersOwnerPat>().NotNull()
```

ReSharper.FSharp/src/FSharp.Psi.Intentions/src/Intentions/UseNamedAccessAction.fs
```diff
@@ -0,0 +1,107 @@
...
+open JetBrains.ReSharper.Psi.ExtensionsAPI
+open JetBrains.ReSharper.Psi.ExtensionsAPI.Tree
+open JetBrains.ReSharper.Resources.Shell
+open JetBrains.ReSharper.Plugins.FSharp.Psi
+
+[<ContextAction(Name = "UseNamedAccess", Group = "F#", Description = "Use named access inside a DU pattern")>]
+type UseNamedAccessAction(dataProvider: FSharpContextActionDataProvider) =
+    inherit FSharpContextActionBase(dataProvider)
+    
+    let mutable names = Array.empty
```
nojaf, line 15: @auduchinok this most likely is a major sacrilege.
What is a better way to deal with the case where you need to re-use nodes that you needed for `IsAvailable` and `ExecutePsiTransaction`? Have them as let bindings in the type?

ReSharper.FSharp/src/FSharp.Psi.Intentions/src/Intentions/UseNamedAccessAction.fs
```diff
@@ -0,0 +1,107 @@
...
+                | _ -> Some (name, pat))
+            |> Seq.toArray
+        
+        let factory = pattern.CreateElementFactory()
+        let sourceText =
+            let fields =
+                usedFieldsWithPatterns
+                |> Array.map (fun (name,_) -> $"{name} = _")
+                |> String.concat "; "
+            $"{pattern.Identifier.Name}({fields})"
```
auduchinok, line 82: This may in theory break multiline parameter owner patterns or multiline field patterns. The tree would be fixed automatically if we had R# formatter finished, but for now we could, perhaps, disable the action on multiline patterns.

ReSharper.FSharp/src/FSharp.Psi.Intentions/src/Intentions/UseNamedAccessAction.fs
```diff
@@ -0,0 +1,107 @@
...
+        let namedPattern = factory.CreatePattern(sourceText, false)
+        
+        match namedPattern with
+        | :? IParametersOwnerPat as ownerPat ->
+            if ownerPat.Parameters.Count <> 1 then null else
```
auduchinok, line 87: Please assert, this has been checked in `IsAvailable` already.

auduchinok (commented):
@nojaf Very nice attention to details throughout the PR, and I do like the idea behind it!

Actually, I was going to suggest you to implement something like this, as I remembered that you liked to use named patterns in your code. 🙂

ReSharper.FSharp/src/FSharp.Psi.Intentions/src/Intentions/UseNamedAccessAction.fs
```diff
@@ -0,0 +1,107 @@
...
+            |> Seq.toArray
+        
+        // All fields need to be named
+        if names.Length <> fcsUnionCase.Fields.Count then false else
+        // The amount of fields needs to match with the tuple length
+        if names.Length <> tuplePat.Patterns.Count then false else
+
+        true
+
+    override x.ExecutePsiTransaction(_,_) =
```
auduchinok, line 61: There's also another overload that you can override: it takes one less parameter and returns `unit` instead of an action. Using that one would allow you to remove all the `null` returning code here.

ReSharper.FSharp/src/FSharp.Psi.Intentions/src/Intentions/UseNamedAccessAction.fs
auduchinok: It's actually the only way I know of doing it. 🙂

ReSharper.FSharp/src/FSharp.Psi.Intentions/src/Intentions/UseNamedAccessAction.fs
nojaf: Updated

ReSharper.FSharp/src/FSharp.Psi.Intentions/src/Intentions/UseNamedAccessAction.fs
nojaf: Added an extra check to not make the intention available when the paren pattern is multiline.

ReSharper.FSharp/src/FSharp.Psi.Intentions/src/Intentions/UseNamedAccessAction.fs
nojaf: Very good feedback here! I guess we should have some sort of heuristic.
If the number of wildcards and constants exceeds a certain threshold, we should propose the named access intention.

What about if 50% of the patterns are either constant or wildcards then propose the intention.
Or always if three or more constant or wildcards.
```fsharp
match x with
| A(a, _{on}) -> () // 50%
| B(a, b, _{off}) -> ()  // 33%
| C(true, true, a, b, d, c, e, _{on} ) // 60% but 3 items can be replaced
```

nojaf:
image

`Deconstruct` action is still above `Use named access`.

ReSharper.FSharp/src/FSharp.Psi.Intentions/src/Intentions/UseNamedAccessAction.fs
nojaf: We settled on: always enabling for now and improving later.

ReSharper.FSharp/src/FSharp.Psi.Intentions/src/Intentions/UseNamedAccessAction.fs
```diff
@@ -0,0 +1,103 @@
...
+        let pattern = dataProvider.GetSelectedElement<IParametersOwnerPat>()
+        if isNull pattern then false else
+        // We expect a single IParenPat here
+        if pattern.Parameters.Count <> 1 then false else
+        let parenPat =
+            match pattern.Parameters.[0] with
+            | :? IParenPat as parenPat -> parenPat
+            | _ -> null
+
+        if isNull parenPat then false else
```
auduchinok, line 32:
```fsharp
let parenPat = pattern.ParametersEnumerable.SingleItem.As<IParenPat>()
if isNull parenPat then false else
```

ReSharper.FSharp/src/FSharp.Psi.Intentions/src/Intentions/UseNamedAccessAction.fs
```diff
@@ -0,0 +1,103 @@
...
+
+        // Ignore multiline patterns for now
+        if parenPat.StartLine <> parenPat.EndLine then false else
+
+        // This should only work when there are multiple items
+        tuplePat <- parenPat.Pattern :?> ITuplePat
+        if isNull tuplePat then false else
+
+        // We need to be sure that the pattern is a DU with all named fields
+        let fcsUnionCase = pattern.ReferenceName.Reference.GetFcsSymbol().As<FSharpUnionCase>()
```
auduchinok, line 42: In theory `pattern.ReferenceName` might be `null` in case of some broken code.

The reference name may also be somewhat broken, but we shouldn't fail in that case too. Please add a test case showing that the context action isn't available here (meaning there're no exceptions produced):
```fsharp
match x with
| Namespace. (a, b) -> ()
```

ReSharper.FSharp/src/FSharp.Psi.Intentions/src/Intentions/UseNamedAccessAction.fs
```diff
@@ -0,0 +1,103 @@
...
+            (names, tuplePat.PatternsEnumerable)
+            ||> Seq.zip
+            |> Seq.choose (fun (name, pat) -> visit pat |> Option.map (fun pat -> name, pat))
+            |> Seq.toArray
+
+        let factory = pattern.CreateElementFactory()
+        let sourceText =
+            let fields =
+                usedFieldsWithPatterns
+                |> Array.map (fun (name,_) -> $"{name} = _")
```
auduchinok, line 80: Is there a Fantomas setting that controls the spaces around `=`?

ReSharper.FSharp/test/src/FSharp.Intentions.Tests/src/Intentions/UseNamedAccessTest.fs
```diff
@@ -0,0 +1,20 @@
...
+    [<Test>] member x.``WildCard 01``() = x.DoNamedTest()
+    [<Test>] member x.``Constant 01``() = x.DoNamedTest()
+    [<Test>] member x.``Tuple 01``() = x.DoNamedTest()
+
+type UseNamedAccessAvailabilityTest() =
+    inherit FSharpContextActionAvailabilityTestBase<UseNamedAccessAction>()
+
+    override x.ExtraPath = "useNamedAccess"
+
+    [<Test>] member x.``Multiline Pattern 01``() = x.DoNamedTest()
```
auduchinok, line 20: Would also be nice to have a test for fields without names.

auduchinok (approved):
@nojaf Good work, just a few suggestions from me.

ReSharper.FSharp/src/FSharp.Psi.Intentions/src/Intentions/UseNamedAccessAction.fs
```diff
@@ -0,0 +1,103 @@
...
+open type JetBrains.Diagnostics.Assertion
+
+[<ContextAction(Name = "UseNamedAccess", Group = "F#", Description = "Use named access inside a DU pattern")>]
+type UseNamedAccessAction(dataProvider: FSharpContextActionDataProvider) =
+    inherit FSharpContextActionBase(dataProvider)
+    
+    let mutable names = Array.empty
+    let mutable tuplePat : ITuplePat = null
+
+    override x.Text = "Use named access"
```
auduchinok, line 20: It should maybe somehow include the union case name. It could help when there are nested patterns.
Or it should use the terminology from the language docs, which says 'named fields'.

It's difficult to combine these two things. 🙂

auduchinok: Maybe something like this (for now), even though it's not the prettiest title I've suggested: `"Use named fields for {unionCase.Name}"`

ReSharper.FSharp/src/FSharp.Psi.Intentions/src/Intentions/UseNamedAccessAction.fs
nojaf: No, the style guide mentions spaces explicitly ([source](https://learn.microsoft.com/en-us/dotnet/fsharp/style-guide/formatting#formatting-named-arguments))

ReSharper.FSharp/src/FSharp.Psi.Intentions/src/Intentions/UseNamedAccessAction.fs
nojaf: `IParametersOwnerPat` is already null in this example.

nojaf:
```
Language: PsiLanguageType:F#
IFSharpImplFile
  IAnonModuleDeclaration
    IExpressionStatement
      IChameleonExpression
        IMatchExpr
          FSharpTokenType+MatchTokenElement(type:MATCH, text:match)
          Whitespace(type:WHITE_SPACE, text: ) spaces:" "
          IReferenceExpr
            FSharpIdentifierToken(type:IDENTIFIER, text:x)
          Whitespace(type:WHITE_SPACE, text: ) spaces:" "
          FSharpTokenType+WithTokenElement(type:WITH, text:with)
          NewLine(type:NEW_LINE, text:\n) spaces:"\n"
          IMatchClause
            FSharpTokenType+BarTokenElement(type:BAR, text:|)
            Whitespace(type:WHITE_SPACE, text: ) spaces:" "
            ILocalReferencePat
              IExpressionReferenceName
                IExpressionReferenceName
                  FSharpIdentifierToken(type:IDENTIFIER, text:Namespace)
                FSharpTokenType+DotTokenElement(type:DOT, text:.)
                Whitespace(type:WHITE_SPACE, text: ) spaces:" "
                FSharpTokenType+LparenTokenElement(type:LPAREN, text:()
                FSharpIdentifierToken(type:IDENTIFIER, text:a)
                FSharpTokenType+CommaTokenElement(type:COMMA, text:,)
                Whitespace(type:WHITE_SPACE, text: ) spaces:" "
                FSharpIdentifierToken(type:IDENTIFIER, text:b)
                FSharpTokenType+RparenTokenElement(type:RPAREN, text:))
            Whitespace(type:WHITE_SPACE, text: ) spaces:" "
            FSharpTokenType+RarrowTokenElement(type:RARROW, text:->)
            Whitespace(type:WHITE_SPACE, text: ) spaces:" "
            IUnitExpr
              FSharpTokenType+LparenTokenElement(type:LPAREN, text:()
              FSharpTokenType+RparenTokenElement(type:RPAREN, text:))
              ```

File: ReSharper.FSharp/src/FSharp.Psi.Intentions/src/Intentions/UseNamedAccessAction.fs (hunk 12 above)
Comment (nojaf) [reply ReSharper.FSharp/src/FSharp.Psi.Intentions/src/Intentions/UseNamedAccessAction.fs]: Done!

**Comment by nojaf** [comment]
Thanks for the review! Great pointers.

File: ReSharper.FSharp/src/FSharp.Psi.Intentions/src/Intentions/UseNamedAccessAction.fs (hunk 9 above)
Comment (auduchinok) [reply ReSharper.FSharp/src/FSharp.Psi.Intentions/src/Intentions/UseNamedAccessAction.fs:41]: > IParametersOwnerPat is already null in this example.

I can see it changing, though. 🙂

Comment (nojaf) [reply ReSharper.FSharp/src/FSharp.Psi.Intentions/src/Intentions/UseNamedAccessAction.fs:41]: Oh, I did add the check for `pattern.ReferenceName` being `null``.
The current test just doesn't cover it.

File: ReSharper.FSharp/src/FSharp.Psi.Intentions/src/Intentions/UseNamedAccessAction.fs [hunk 13]
```diff
@@ -17,7 +17,13 @@ type UseNamedAccessAction(dataProvider: FSharpContextActionDataProvider) =
     let mutable names = Array.empty
     let mutable tuplePat : ITuplePat = null
 
-    override x.Text = "Use named access"
+    override x.Text = 
+        let pattern = dataProvider.GetSelectedElement<IParametersOwnerPat>()
+        if isNull pattern then
+            "Use named fields in pattern"
```
auduchinok, line 23: I think the text will only be accessed when `IsAvailable` returned `true`, so this pattern should never be `null` after that.

ReSharper.FSharp/src/FSharp.Psi.Intentions/src/Intentions/UseNamedAccessAction.fs
```diff
@@ -0,0 +1,108 @@
...
+    let mutable names = Array.empty
+    let mutable patterns : IFSharpPattern seq = Seq.empty
+
+    override x.Text = 
+        let pattern = dataProvider.GetSelectedElement<IParametersOwnerPat>()
+        if isNull pattern then
+            "Use named fields in pattern"
+        else
+            let unionCaseName = pattern.ReferenceName.Names |> String.concat "."
+            $"Use named fields for '{unionCaseName}'"
```
auduchinok, line 26:
```suggestion
            $"Use named fields for '{pattern.ReferenceName.ShortName}'"
```

ReSharper.FSharp/src/FSharp.Psi.Intentions/src/Intentions/UseNamedAccessAction.fs
nojaf: Yeah wasn't sure about that part.

auduchinok:
@nojaf congrats and thanks! 🙂

PR #502, opened 2023-03-22
remove screenshots from the "dump psi to file" documentation

dawedawe:
As discussed on slack, this removes the screenshots from the "dump psi to file" documention.


auduchinok (approved):
Thanks!

PR #504, opened 2023-03-29
CodeStyle Settings page: support .editorconfig layer

DedSec256:
Fix for [RIDER-91470](https://youtrack.jetbrains.com/issue/RIDER-91470/There-is-no-descriptor-for-an-options-page-ID-FSharpCodeStylePage)


auduchinok (approved):
Thanks!

PR #512, opened 2023-04-14
improve formatting of members and fields when generating overrides

dawedawe:
Before:
```fsharp
type R = 
  {
  F: int
  G: int
  }

  override this.ToString() = failwith "todo"
```

After:
```fsharp
type R = 
  {
    F: int
    G: int
  }

  override this.ToString() = failwith "todo"
```

Before:
```fsharp
type C() = 
  class
  override this.ToString() = failwith "todo"

  member this.X = [ 1
           2
           3 ]
  end
```

After:
```fsharp
type C() = 
  class
    override this.ToString() = failwith "todo"

    member this.X = [ 1
                      2
                      3 ]
  end
```


ReSharper.FSharp/src/FSharp.Psi.Services/src/Generate/GenerateProvider.fs
```diff
@@ -200,26 +200,51 @@ type FSharpOverridingMembersBuilder() =
...
-                    shiftNode -diff m
+            let diff = m.Indent - desiredIndent
+            if diff > 0 then shiftWithWhitespaceBefore -diff m)
+
+        // members between representation start/end might need further indentation
+        typeRepr.GetMemberDeclarations()
+        |> Seq.iter (fun m ->
+            match m with
+            | :? IRecordFieldDeclaration
+            | :? IMemberDeclaration when typeRepr.Contains(m) ->
```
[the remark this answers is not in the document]
auduchinok: Is there a case where `typeRepr.Contains(m)` is `false` for a member `m` that is coming from `typeRepr.GetMemberDeclarations()`?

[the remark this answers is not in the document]
dawedawe: > Is there a case where `typeRepr.Contains(m)` is `false` for a member `m` that is coming from `typeRepr.GetMemberDeclarations()`?

No, that was overly cautious by me.

There's still the issue that overrides are added before let bindings but I'd prefer to handle that in a separate PR as it is anchor related and not about indentation and formatting.

ReSharper.FSharp/test/data/features/generate/overrides/Anchor - Repr 13.fs.gold
```diff
@@ -0,0 +1,16 @@
...
+
+// ${KIND:Overrides}
+// ${SELECT0:ToString():System.String}
+
+// ${KIND:Overrides}
+
+type C() = 
+  class
+    override this.ToString() = {selstart}failwith "todo"{selend}
+    let f = 23
```
auduchinok, line 15: This order is incorrect.

ReSharper.FSharp/src/FSharp.Psi.Services/src/Generate/GenerateProvider.fs
```diff
@@ -198,28 +198,75 @@ type FSharpOverridingMembersBuilder() =
...
+        declarations
+        |> Seq.iter (fun m ->
+            match m with
+            | :? IValFieldDeclaration
+            | :? IAbstractMemberDeclaration
+            | :? IConstructorDeclaration
+            | :? IRecordFieldDeclaration
+            | :? ILetBindingsDeclaration
+            | :? IInterfaceImplementation
+            | :? IMemberDeclaration ->
```
auduchinok, line 266: It's been already discussed in the previous PRs, please try to use base interfaces instead of more specific ones where possible.

ReSharper.FSharp/test/data/features/generate/overrides/Anchor - Repr 13.fs.gold
auduchinok: I see that it's going to be addressed separately (https://github.com/JetBrains/resharper-fsharp/pull/512#discussion_r1169020142).

ReSharper.FSharp/src/FSharp.Psi.Services/src/Generate/GenerateProvider.fs
dawedawe: Ah yeah, sorry. Got carried away there.

ReSharper.FSharp/src/FSharp.Psi.Services/src/Generate/GenerateProvider.fs
```diff
@@ -198,28 +198,71 @@ type FSharpOverridingMembersBuilder() =
...
+        match typeRepr with
+        | :? IClassRepresentation as classRepr ->
+            normalizeReprEnd classRepr.BeginKeyword classRepr.EndKeyword
+        | :? IStructRepresentation as structRepr ->
+            normalizeReprEnd structRepr.BeginKeyword structRepr.EndKeyword
+        | :? IRecordRepresentation as recordRepr ->
+            normalizeReprEnd recordRepr.LeftBrace recordRepr.RightBrace
+
+            if recordRepr.FieldDeclarations.Count > 1 then
+                recordRepr.FieldDeclarations
```
auduchinok, line 225: Please extract `recordRepr.FieldDeclarations` to a binding, it's being calculated 4 times in this `if` expression.

ReSharper.FSharp/src/FSharp.Psi.Services/src/Generate/GenerateProvider.fs
```diff
@@ -198,28 +198,71 @@ type FSharpOverridingMembersBuilder() =
...
+                |> Seq.skip 1
+                |> Seq.iter (fun (idx, f) ->
+                    let prevField = recordRepr.FieldDeclarations[idx - 1]
+                    if f.StartLine <> prevField.StartLine && f.Indent > recordRepr.FieldDeclarations[0].Indent then
+                        let diff = f.Indent - recordRepr.FieldDeclarations[0].Indent
+                        shiftWithWhitespaceBefore -diff f)
+        | _ -> ()
+
+        // members might need deindenting after the representation start was moved to it's own line
         typeDecl.TypeMembers
```
auduchinok, line 236: It seems that moving members only may produce unexpected result in cases like this one:
```fsharp
type R = { F: int }
          // comment
          member this.P = 1
```

ReSharper.FSharp/src/FSharp.Psi.Services/src/Generate/GenerateProvider.fs
```diff
@@ -212,6 +212,9 @@ type FSharpOverridingMembersBuilder() =
             if beginToken.Indent > endToken.Indent then
                 let diff = beginToken.Indent - endToken.Indent
                 addNodeBefore endToken (Whitespace(diff))
+            else if endToken.Indent > beginToken.Indent && isFirstMeaningfulNodeOnLine endToken then
```
auduchinok, line 215:
```suggestion
            elif endToken.Indent > beginToken.Indent && isFirstMeaningfulNodeOnLine endToken then
```

ReSharper.FSharp/src/FSharp.Psi.Services/src/Generate/GenerateProvider.fs
```diff
@@ -230,38 +233,34 @@ type FSharpOverridingMembersBuilder() =
...
+            nodes
+            |> Seq.iter (fun m ->
+                let diff = m.Indent - indentationTarget
+                if diff > 0  && isFirstMeaningfulNodeOnLine m then shiftWithWhitespaceBefore -diff m)
+
+        // members etc. might need indenting or deindenting after the representation start was moved to it's own line
+        let reprNodes =
+            typeRepr.Children()
+            |> Seq.filter
+                   (fun c -> c :? ICommentNode || c :? ITypeBodyMemberDeclaration || c :? IRecordFieldDeclarationList)
```
auduchinok, line 246: Does it need to be specific about the node types? Maybe it could skip `{`/`class`/`struct` and move any nodes until `}`/`end`?

ReSharper.FSharp/src/FSharp.Psi.Services/src/Generate/GenerateProvider.fs
```diff
@@ -230,38 +233,34 @@ type FSharpOverridingMembersBuilder() =
...
+        deindent (desiredIndent + indentSize) reprNodes
+
+        let declNodes =
+            typeDecl.Children() |> Seq.filter (fun c -> c :? ICommentNode || c :? ITypeBodyMemberDeclaration)
```
auduchinok, line 250: The same, maybe just move all nodes after the representation one?

ReSharper.FSharp/src/FSharp.Psi.Services/src/Generate/GenerateProvider.fs
```diff
@@ -236,15 +238,28 @@ type FSharpOverridingMembersBuilder() =
...
+                children[first .. last] |> List.filter (isWhitespace >> not)
+            | _ -> List.empty
         deindent (desiredIndent + indentSize) reprNodes
 
         let declNodes =
-            typeDecl.Children() |> Seq.filter (fun c -> c :? ICommentNode || c :? ITypeBodyMemberDeclaration)
+            let children = typeDecl.Children() |> Seq.toList
+            let idx = children |> List.tryFindIndex(fun c -> c :? ITypeRepresentation)
+            match idx with
+            | Some idx when idx >= 0 -> List.skip (idx + 1) children |> List.filter (isWhitespace >> not)
```
auduchinok, line 261: `Seq.skipWhile (function | :? ITypeRepresentation -> false | _ -> true)`
or
`Seq.skipWhile ((!=) typeRepr)`

auduchinok: > `List.filter (isWhitespace >> not)`

I think you should also update the whitespace nodes, at least on lines where whitespaces are the only nodes on a line.

There's a whitespace of length 12 on the line between the fields here:
```fsharp
type R = { F1: int
           
           F2: int }
```

It's not being updated here:
```fsharp
type R =
    { F1: int
           
      F2: int }
```

ReSharper.FSharp/src/FSharp.Psi.Services/src/Generate/GenerateProvider.fs
```diff
@@ -236,15 +238,28 @@ type FSharpOverridingMembersBuilder() =
...
 
-        // members etc. might need indenting or deindenting after the representation start was moved to it's own line
         let reprNodes =
-            typeRepr.Children()
-            |> Seq.filter
-                   (fun c -> c :? ICommentNode || c :? ITypeBodyMemberDeclaration || c :? IRecordFieldDeclarationList)
+            let children = typeRepr.Children() |> Seq.toList
+            let startIdx = children |> List.tryFindIndex (fun c -> c.NodeType = FSharpTokenType.CLASS
+                                                                    || c.NodeType = FSharpTokenType.STRUCT
+                                                                    || c.NodeType = FSharpTokenType.LBRACE)
```
auduchinok, line 245: I think it could be easier to copy the approach you've used for normalizing the representations, where you get `beginToken` and `endToken`. It should be possible to reuse that.

```fsharp
children
|> Seq.skipWhile ((!=) beginToken)
|> Seq.tail
|> Seq.takeWhile ((!=) endToken) // assuming there's a `null` check for `endToken`
```

auduchinok: Another approach could be this:
```fsharp
TreeRange(beginToken.NextSibling, endToken.PrevSibling)
```

ReSharper.FSharp/src/FSharp.Psi.Services/src/Generate/GenerateProvider.fs
dawedawe: Okay, as the empty line was handled as part of the parent node I introduced some traversing to get to the empy-line children.

ReSharper.FSharp/src/FSharp.Psi.Services/src/Generate/GenerateProvider.fs
```diff
@@ -195,29 +196,91 @@ type FSharpOverridingMembersBuilder() =
...
+                (fun n ->
+                    if isAtEmptyLine n then
+                        let nextToken =  n.GetNextToken()
+                        if isNotNull nextToken then
+                            let diff = nextToken.Indent - indentationTarget
+                            match n with
+                            | :? Whitespace as whitespace -> if diff > 0 then shiftWhitespaceBefore -diff whitespace
+                            | _ -> ()
+                    else
+                        deindentEmptyLines indentationTarget (n.Children()))
```
auduchinok, line 250: @dawedawe It looks like a wrong place to do it, and it adds a lot of complication to the code. Have you considered updating `shiftNode` or `shiftWithWhitespaceBefore`, so there's no need to process some of the tokens separately in every feature? It seems like it should be a ~two line fix somewhere in there instead of ~twenty+ lines of special checks here.

dawedawe: I thought it wouldn't be acceptable to change such an infrastructure code, so I shied away from it.
But if you're okay with it, all the better. There's no impact on other existing tests.

To let existing shifting functions work on the representation nodes so that whitespace in empty lines is also shifted, I still had to rearrange some things...
I think it works now as you want for representations, but not for empty lines which are just in the declaration.
Given that many people use `trim_trailing_whitespace` in their editorconfig, I'm not sure how much effort we should put into it. But, of course, it's your decision.

auduchinok: > I thought it wouldn't be acceptable to change such an infrastructure code, so I shied away from it.
> But if you're okay with it, all the better. There's no impact on other existing tests.

It's more than welcome to improve it. 🙂

dawedawe:
Ok, I was able to make the desired change and keep the tests green.
But as the typeRepr.Indent has changed I had to make quite some changes to diff calculations and shifting operations.
Let me know, if there's a better way to do this.

ReSharper.FSharp/src/FSharp.Psi.Services/src/Generate/GenerateProvider.fs
```diff
@@ -199,18 +199,17 @@ type FSharpOverridingMembersBuilder() =
         let indentSize = typeDecl.GetIndentSize()
         let desiredIndent = typeDecl.Indent + indentSize
 
-        addNodesBefore typeRepr.FirstChild [
-            NewLine(typeRepr.GetLineEnding())
-            Whitespace(typeRepr.Indent)
-        ] |> ignore
+        let origTypeReprIndent = typeRepr.Indent
+        addNodeBefore typeRepr (NewLine(typeRepr.GetLineEnding()))
+        addNodeBefore typeRepr.FirstChild (Whitespace(origTypeReprIndent))
```
auduchinok, line 204: This still (likely) adds a child to a wrong place. Please check the tree structure with a parser test.

Why likely? There's some logic in R# that tries to move whitespace nodes (e.g. `Whitespace`, `NewLine`) to parent nodes when they are at a node boundary (i.e. a node shouldn't start/end with a whitespace). So there's a chance that this logic fixes the tree structure for you, but the modification here itself looks wrong.

ReSharper.FSharp/src/FSharp.Psi.Services/src/Generate/GenerateProvider.fs
```diff
@@ -256,11 +255,15 @@ type FSharpOverridingMembersBuilder() =
             for node in range do
                 if not (isFirstMeaningfulNodeOnLine node) then () else
 
-                let diff = node.Indent - (desiredIndent + additionalIndent)
+                let diff =
+                    if node = typeRepr then
```
auduchinok, line 259: Please use `==`, we're interested in reference equality only when working with these tree nodes.

ReSharper.FSharp/src/FSharp.Psi.Services/src/Generate/GenerateProvider.fs
auduchinok: The dump looks correct, all seems good now!

auduchinok (approved):
@dawedawe Very nice work, thanks!

PR #531, opened 2023-05-25
Language injections: fix backslash escaping at the end of the string part

DedSec256:
Currently, we don't escape backslashes at the end of the string part (regular or regular interpolated) that will lead to a parsing error in the corresponding F# string. 

Correct escaping examples:

Language injections

- [x] Tests


auduchinok (approved):
Thanks for adding the tests!

PR #536, opened 2023-06-16
Fix compile after INTRA_TEXT adornments api changes

nickplus11:


auduchinok (approved):
@nickplus11 Thanks a lot!

PR #542, opened 2023-07-05
Use WRAP_LIMIT to limit the width of tooltip signatures

dawedawe:
This makes use of https://github.com/dotnet/fsharp/pull/14891 to squash tooltip signatures if they are longer than WRAP_LIMIT


dawedawe:
And here's the division by 3.
image

auduchinok:
@dawedawe Could you please rebase to `net233`?

PR #543, opened 2023-07-05
Quick fix for ValueNotContainedMutabilityAttributesDiffer

nojaf:
We are interested in providing some quick fixes for `FS0034`.

See [FSComp.txt](https://github.com/dotnet/fsharp/blob/723b1dc6310d773070804c0d258b2227c77d1ef8/src/Compiler/FSComp.txt#L102-L125)

ValueNotContainedMutabilityAttributesDiffer

Eventually, we would hope to address the 'Type mismatch' errors once the `GetValSignatureText` API is available. In the meantime, we can start with the smaller ones, like this PR.

### DIFF
--- a/ReSharper.FSharp/src/FSharp.Psi.Intentions/src/QuickFixes/UpdateMutabilityInSignatureFix.fs
+++ b/ReSharper.FSharp/src/FSharp.Psi.Intentions/src/QuickFixes/UpdateMutabilityInSignatureFix.fs
@@ -1,4 +1,4 @@
-﻿namespace JetBrains.ReSharper.Plugins.FSharp.Psi.Features.Daemon.QuickFixes
+namespace JetBrains.ReSharper.Plugins.FSharp.Psi.Features.Daemon.QuickFixes
 
 open JetBrains.ReSharper.Plugins.FSharp.Psi.Tree
 open JetBrains.ReSharper.Plugins.FSharp.Psi.Features.Daemon.Highlightings
@@ -53,7 +53,7 @@ type UpdateMutabilityInSignatureFix(error: ValueNotContainedMutabilityAttributes
     override x.IsAvailable _ =
         match tryFindImplementationBindingInfo error.Pat with
         | None -> false
-        |Some (topLevelBinding, declaredElement) ->
+        | Some (topLevelBinding, declaredElement) ->
             match tryFindBindingSignature error.Pat.DeclaredName declaredElement with
             | None -> false
             | Some bindingSignature -> topLevelBinding.IsMutable <> bindingSignature.IsMutable
@@ -64,7 +64,7 @@ type UpdateMutabilityInSignatureFix(error: ValueNotContainedMutabilityAttributes
 
         match tryFindImplementationBindingInfo error.Pat with
         | None -> ()
-        |Some (topLevelBinding, declaredElement) ->
+        | Some (topLevelBinding, declaredElement) ->
             match tryFindBindingSignature error.Pat.DeclaredName declaredElement with
             | None -> ()
             | Some bindingSignature ->
--- a/ReSharper.FSharp/src/FSharp.Psi.Intentions/src/QuickFixes/UpdateMutabilityInSignatureFix.fs
+++ b/ReSharper.FSharp/src/FSharp.Psi.Intentions/src/QuickFixes/UpdateMutabilityInSignatureFix.fs
@@ -1,0 +1,71 @@
+﻿namespace JetBrains.ReSharper.Plugins.FSharp.Psi.Features.Daemon.QuickFixes
+
+open JetBrains.ReSharper.Plugins.FSharp.Psi.Tree
+open JetBrains.ReSharper.Plugins.FSharp.Psi.Features.Daemon.Highlightings
+open JetBrains.ReSharper.Psi.ExtensionsAPI
+open JetBrains.ReSharper.Resources.Shell
+open JetBrains.ReSharper.Plugins.FSharp.Psi
+
+type UpdateMutabilityInSignatureFix(error: ValueNotContainedMutabilityAttributesDifferError) =
+    inherit FSharpQuickFixBase()
+
+    let tryFindBindingSignature (implementationDeclaredName: string) (declaredElement: IFSharpMember) =
+        let parentDeclarations = declaredElement.ContainingType.GetDeclarations()
+
+        // Find the parent signature counter part
+        let signatureNameModuleDeclaration =
+            parentDeclarations
+            |> Seq.tryPick (fun d ->
+                match d with
+                | :? INamedModuleDeclaration as signatureModule when
+                    signatureModule.GetSourceFile().IsFSharpSignatureFile -> Some signatureModule
+                | _ -> None)
+
+        signatureNameModuleDeclaration
+        |> Option.bind (fun signatureModule ->
+            signatureModule.Members
+            |> Seq.tryPick (fun sigMember ->
+                match sigMember with
+                | :? IBindingSignature as bindingSignature ->
+                    match bindingSignature.HeadPattern with
+                    | :? ITopReferencePat as bindingNamePat
+                        // This is a bit lucky to work.
+                        // There must be something more clever we can try
+                        when bindingNamePat.DeclaredName = implementationDeclaredName ->
+                        Some bindingSignature
+                    | _ -> None
+                | _ -> None)
+        )
+
+    let tryFindImplementationBindingInfo (pat: ITopReferencePat) =
+        if isNull pat then None else
+        let topLevelBinding = TopBindingNavigator.GetByHeadPattern(pat)
+        if isNull topLevelBinding then None else
+        let declaredElement = topLevelBinding.DeclaredElement :?> IFSharpMember
+        if isNull declaredElement then None else
+        let mfv = declaredElement.Mfv
+        if isNull mfv then None else
+        if not mfv.HasSignatureFile then None else
+        Some (topLevelBinding, declaredElement)
+
+    override x.Text = "Update mutability in signature"
+
+    override x.IsAvailable _ =
+        match tryFindImplementationBindingInfo error.Pat with
+        | None -> false
+        |Some (topLevelBinding, declaredElement) ->
+            match tryFindBindingSignature error.Pat.DeclaredName declaredElement with
+            | None -> false
+            | Some bindingSignature -> topLevelBinding.IsMutable <> bindingSignature.IsMutable
+
+    override x.ExecutePsiTransaction _ =
+        use writeCookie = WriteLockCookie.Create(error.Pat.IsPhysical())
+        use disableFormatter = new DisableCodeFormatter()
+
+        match tryFindImplementationBindingInfo error.Pat with
+        | None -> ()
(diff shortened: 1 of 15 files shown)


ReSharper.FSharp/src/FSharp.Psi/src/Impl/Tree/BindingSignature.cs
```diff
@@ -8,8 +9,18 @@ internal partial class BindingSignature
 
     public void SetIsMutable(bool value)
     {
-      if (!value)
-        throw new System.NotImplementedException();
+      if (!value && MutableKeyword != null)
+      {
+        if (MutableKeyword.PrevSibling is Whitespace whitespace)
```
auduchinok, line 14: We can also try not disabling the formatter for this change in a signature file, and then, in theory, you won't have to deal with the spaces here.

ReSharper.FSharp/src/FSharp.Psi.Intentions/src/QuickFixes/UpdateMutabilityInSignatureFix.fs
```diff
@@ -0,0 +1,71 @@
...
+    let tryFindImplementationBindingInfo (pat: ITopReferencePat) =
+        if isNull pat then None else
+        let topLevelBinding = TopBindingNavigator.GetByHeadPattern(pat)
+        if isNull topLevelBinding then None else
+        let declaredElement = topLevelBinding.DeclaredElement :?> IFSharpMember
+        if isNull declaredElement then None else
+        let mfv = declaredElement.Mfv
+        if isNull mfv then None else
+        if not mfv.HasSignatureFile then None else
+        Some (topLevelBinding, declaredElement)
```
auduchinok, line 49: `TopBindingNavigator.GetByHeadPattern(pat)` won't cover cases like
```fsharp
let (a, b) = ...
```
or
```fsharp
let (i: int) = ...
```
I'm not sure if we want to cover them, but if we do, there's `Binding` property that makes it easier to get a binding through various nested patterns.

Getting a declared element from a binding would only work reliably for a single top-level pattern, so getting it from the initial pattern would be better too.

This is what the change could look like if want to support nested patterns:
```suggestion
        if isNull pat then None else

        match pat.Binding, pat.DeclaredElement.As<IFSharpMember>() with
        | null, _ | _, null -> None
        | binding, fsMember ->

        let mfv = fsMember.Mfv
        if isNull mfv then None else

        Some(binding, fsMember)
```

The `x.ExecutePsiTransaction` would also get simplified if you use `Binding.

ReSharper.FSharp/src/FSharp.Psi.Intentions/src/QuickFixes/UpdateMutabilityInSignatureFix.fs
```diff
@@ -0,0 +1,71 @@
...
+open JetBrains.ReSharper.Plugins.FSharp.Psi.Features.Daemon.Highlightings
+open JetBrains.ReSharper.Psi.ExtensionsAPI
+open JetBrains.ReSharper.Resources.Shell
+open JetBrains.ReSharper.Plugins.FSharp.Psi
+
+type UpdateMutabilityInSignatureFix(error: ValueNotContainedMutabilityAttributesDifferError) =
+    inherit FSharpQuickFixBase()
+
+    let tryFindBindingSignature (implementationDeclaredName: string) (declaredElement: IFSharpMember) =
+        let parentDeclarations = declaredElement.ContainingType.GetDeclarations()
```
auduchinok, line 13: It should be possible to just do `declaredElement.GetDeclarations`, unless there's some issue. It would return other patterns declaring this element.

ReSharper.FSharp/src/FSharp.Psi.Services/src/Daemon/Highlightings/FcsErrors.xml
```diff
@@ -207,6 +207,17 @@
     <QuickFix>RemoveUnnecessaryUpcastFix</QuickFix>
   </Warning>
 
+  <Error staticGroup="FSharpErrors" name="ValueNotContainedMutabilityAttributesDiffer" ID="FS0034: but its signature specifies The mutability attributes differ">
```
auduchinok, line 210: It seems the error description got cut off?

ReSharper.FSharp/test/src/FSharp.Intentions.Tests/src/QuickFixes/UpdateMutabilityInSignatureFixTest.fs
```diff
@@ -0,0 +1,12 @@
+﻿namespace JetBrains.ReSharper.Plugins.FSharp.Tests.Intentions.QuickFixes
+
+open JetBrains.ReSharper.Plugins.FSharp.Psi.Features.Daemon.QuickFixes
+open JetBrains.ReSharper.Plugins.FSharp.Tests
+open NUnit.Framework
+
+[<FSharpTest>]
+type UpdateMutabilityInSignatureFixTest() =
+    inherit FSharpQuickFixTestBase<UpdateMutabilityInSignatureFix>()
+    override x.RelativeTestDataPath = "features/quickFixes/updateMutabilityInSignatureFix"
```
auduchinok, line 10: Nit-picking: could you please add empty lines after `inherit` and `override` members?

ReSharper.FSharp/test/src/FSharp.Intentions.Tests/src/QuickFixes/UpdateMutabilityInSignatureFixTest.fs
```diff
@@ -0,0 +1,12 @@
...
+    [<Test>] member x.``Add Mutability - 01`` () = x.DoNamedTestWithSignature()
+    [<Test>] member x.``Remove Mutability - 01`` () = x.DoNamedTestWithSignature()
```
auduchinok, line 12: Could you please add tests covering nested patterns that were discussed above (even if keep the quick fix disabled for them)?

ReSharper.FSharp/src/FSharp.Psi.Intentions/src/QuickFixes/UpdateMutabilityInSignatureFix.fs
```diff
@@ -0,0 +1,71 @@
...
+        let topLevelBinding = TopBindingNavigator.GetByHeadPattern(pat)
+        if isNull topLevelBinding then None else
+        let declaredElement = topLevelBinding.DeclaredElement :?> IFSharpMember
+        if isNull declaredElement then None else
+        let mfv = declaredElement.Mfv
+        if isNull mfv then None else
+        if not mfv.HasSignatureFile then None else
+        Some (topLevelBinding, declaredElement)
+
+    override x.Text = "Update mutability in signature"
```
auduchinok, line 51: We probably want to have the pattern name included here, especially if we want to enable it for nested patterns now or later.

auduchinok:
@nojaf Nice work! It would be cool to be able to update the implementation too, so if you'd wouldn't mind covering it in this or subsequent PRs, it would be great. 🙂

ReSharper.FSharp/test/src/FSharp.Intentions.Tests/src/QuickFixes/UpdateMutabilityInSignatureFixTest.fs
nojaf: Added a test for `(a:int)`. Not sure what other tests I could still add.

ReSharper.FSharp/src/FSharp.Psi/src/Impl/Tree/BindingSignature.cs
```diff
@@ -8,8 +9,18 @@ internal partial class BindingSignature
 
     public void SetIsMutable(bool value)
     {
-      if (!value)
-        throw new System.NotImplementedException();
+      if (!value && MutableKeyword != null)
```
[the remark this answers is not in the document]
nojaf: I'm not 100% sure I got it. Let me know if it still needs a tweak.

ReSharper.FSharp/src/FSharp.Psi.Services/src/Daemon/Highlightings/FcsErrors.xml
nojaf: I updated it. Is my assumption correct that the error specific parts of the message are to be left out of this?

The original error was: `C:\Users\nojaf\Projects\SampleWithSigs\A.fs(3,14): error FS0034: Module 'A' contains↔    val mutable a: int    ↔but its signature specifies↔    val a: int    ↔The mutability attributes differ [C:\Users\nojaf\Projects\SampleWithSigs\SampleWithSigs.fsproj]`

The `val mutable a: int` needs to go right?

ReSharper.FSharp/src/FSharp.Psi/src/Impl/Tree/BindingSignature.cs
auduchinok: Yes, correct. Not 100% safe, but it's where we want to end up for things like this, and we should try enabling it in some simpler cases. Here it only need to process whitespace between two tokens, so I'd try it. 🙂

ReSharper.FSharp/src/FSharp.Psi.Intentions/src/QuickFixes/UpdateMutabilityInSignatureFix.fs
auduchinok: > It doesn't seem to work when the return type of the signature is different.

Yes, for now it's considered to be two different symbols because of the signature mismatch. That's something I'll need to look into when there're more features like this one are added.

ReSharper.FSharp/src/FSharp.Psi/src/Impl/Tree/BindingSignature.cs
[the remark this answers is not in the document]
auduchinok: Yes, all seems right.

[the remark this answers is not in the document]
auduchinok: > (Would also be nice if you updated the second branch too, but it's not strictly necessary as part of this PR)

This one is about `value` being `true`, and we could check if there's a `mutable` already on line 29.

ReSharper.FSharp/src/FSharp.Psi.Intentions/src/QuickFixes/UpdateMutabilityInSignatureFix.fs
nojaf: I've added the name in the pattern. Let me know if you had something else in mind.

ReSharper.FSharp/test/src/FSharp.Intentions.Tests/src/QuickFixes/UpdateMutabilityInSignatureFixTest.fs
nojaf: Added more tests for patterns

nojaf:
Alright, I think this is pretty close.
Let me know if this needs any more tweaks.

auduchinok (approved):
@nojaf Thanks!

PR #548, opened 2023-07-18
Align bindingSignature with pars.fsy grammar rule.

nojaf:
To align with https://github.com/dotnet/fsharp/blob/8eada0e8c6e28f077899b6757dba16f2dcfddbb8/src/Compiler/pars.fsy#L728


auduchinok (approved):
@nojaf Thanks!

PR #549, opened 2023-07-18
Value not contained mutability accessibility more

nojaf:
RiderUpdateAccessibilityInSignature


ReSharper.FSharp/src/FSharp.Psi/src/FSharp.psi
```diff
@@ -1369,6 +1370,8 @@ autoPropertyDeclaration options { stubBase="FSharpProperTypeMemberDeclarationBas
   STATIC<STATIC, StaticKeyword>?
   memberKeyword{MEMBER, MemberKeyword}
   VAL<VAL, ValKeyword>
+  MUTABLE<MUTABLE, MutableKeyword>?
```
nojaf, line 1373: See https://github.com/dotnet/fsharp/blob/8eada0e8c6e28f077899b6757dba16f2dcfddbb8/src/Compiler/pars.fsy#L2153-L2197

PR #562, opened 2023-09-11
RIDER-15832: put custom operations at the top of completion list

jindraivanek:
Fixes https://youtrack.jetbrains.com/issue/RIDER-15832/Put-custom-operations-at-the-top-of-completion-list-if-its-invoked-on-top-level-inside-computation-expression

This PR puts custom operation on top of completion list in positions where custom operation is possible.

Cleaned up version from https://amplifying-fsharp.github.io/sessions/2023/09/08/

In session, test was of `List` type, but it looks like unrelated changes can break it easily - I changed it to `Selection` type. It seems to work nicely here, although I didn't find it's use anywhere else.


ReSharper.FSharp/src/FSharp.Psi.Features/src/CodeCompletion/Rules/FSharpRelevanceRule.fs
```diff
@@ -39,6 +41,18 @@ type FSharpRelevanceRule() =
...
+                        let isCustomOperationPossible =
+                            let reference = context.ReparsedContext.Reference
+                            if isNull reference then false else
+
+                            let refExpr = reference.GetTreeNode().As<IReferenceExpr>()
+                            if isNull refExpr then false else
+
+                            isInsideComputationExpressionForCustomOperation refExpr
+                        if isCustomOperationPossible then
+                            markRelevance item CLRLookupItemRelevance.ExpectedTypeMatch
```
jindraivanek, line 54: `ExpectedTypeMatch` is used only because it is top relevance. Maybe we want add new relevance?

ReSharper.FSharp/src/FSharp.Psi.Services/src/Util/FSharpExpressionUtil.fs
```diff
@@ -196,6 +196,29 @@ let tryGetEffectiveParentComputationExpression (expr: IFSharpExpression) =
 
     loop false expr
 
+let isInsideComputationExpressionForCustomOperation (expr: IFSharpExpression) =
+    let rec loop (expr: IFSharpExpression) =
+        let computationExpr = ComputationExprNavigator.GetByExpression(expr)
+        let appExpr = PrefixAppExprNavigator.GetByArgumentExpression(computationExpr)
+        if isNotNull appExpr && isNotNull computationExpr then true else
+
+        let letOrUseExpr = LetOrUseExprNavigator.GetByInExpression(expr)
```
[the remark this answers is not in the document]
jindraivanek: Actually is not accepted, I added test for that case: https://github.com/JetBrains/resharper-fsharp/pull/562/files#diff-29c5e797598ed494143ff2fb021971acc44f6c13bf41092181118dd99d07bda2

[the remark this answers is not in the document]
nojaf: Great!

ReSharper.FSharp/src/FSharp.Psi.Services/src/Util/FSharpExpressionUtil.fs
```diff
@@ -196,6 +196,29 @@ let tryGetEffectiveParentComputationExpression (expr: IFSharpExpression) =
 
     loop false expr
 
+let isInsideComputationExpressionForCustomOperation (expr: IFSharpExpression) =
```
auduchinok, line 199: A somewhat corner case example is typing a custom operator before an existing argument:

Screenshot 2023-09-18 at 18 06 27

Not critical, but can likely be supported with a single addition check.

ReSharper.FSharp/src/FSharp.Psi.Features/src/CodeCompletion/Rules/FSharpRelevanceRule.fs
```diff
@@ -39,6 +41,18 @@ type FSharpRelevanceRule() =
...
                 else
+                    if fcsLookupItem.FcsSymbolUse.IsFromComputationExpression then
+                        let isCustomOperationPossible =
+                            let reference = context.ReparsedContext.Reference
+                            if isNull reference then false else
+
+                            let refExpr = reference.GetTreeNode().As<IReferenceExpr>()
+                            if isNull refExpr then false else
+
+                            isInsideComputationExpressionForCustomOperation refExpr
```
auduchinok, line 52: It would be nice to calculate it once if there're many custom operators, so we don't try to traverse the tree for each one again. On the other hand, it looks like the traversal should be short enough, so I'm not sure if this can ever be an actual issue. We could either use something like `lazy` outside the loop, or store the data in the completion context, like `context.GetOrCreateData` (see its other usages in the plugin), if we want to improve it.

jindraivanek: Done

jindraivanek:
Thanks for review, @auduchinok. 

I fixed all except "before argument", that seems more complicated, maybe custom oper.  is not in the list. I'll take closer look later.

auduchinok (approved):
Thanks @jindraivanek, looks good to go in!

nojaf:
Great work @jindraivanek, I'm very happy we did this 🤘

jindraivanek:
Happy how both Amplifying F# session and this PR turns out! 😄

PR #577, opened 2023-11-20
FSharpTreeBuilderBase: support hash type usage

DedSec256:


auduchinok (approved):
@DedSec256 Good fix and great testing, thanks!

PR #585, opened 2023-12-13
LambdaAnalyzer: improves for built-in functions

DedSec256:
Do not suggest warning without quickfix for a built-in function


auduchinok (approved):
Cool!

PR #595, opened 2024-02-14
Experiment: transparent

nojaf:
Trying to use the transparent compiler in Rider.


nojaf:
To start somewhere I would start by using the new `FSharpSnapshotProject` and use the new API.
When `FSharpChecker.Create(useTransparentCompiler = false)` is should just call the old methods anyway: https://github.com/dotnet/fsharp/blob/02d11c67bf156bb406c127d3a552352f8696927a/src/Compiler/Service/BackgroundCompiler.fs#L1631-L1646

We will need to check of course if that is always the case.

PR #620, opened 2024-04-30
Fantomas: Check if .editorconfig overrides option defaults

DedSec256:
Fix for https://youtrack.jetbrains.com/issue/RIDER-111743/Reformat-Code-action-adds-final-newline-in-F-files-against-.editorconfig-setting

- [x] Tests


auduchinok (approved):

DedSec256:
Waiting for some infrastructure problems associated with the transition to the 2.0 plugin model to be fixed so that we can write tests

PR #700, opened 2024-08-16
Fix GitHub Actions 

DedSec256:
- [x] Check targets instead of props


DedSec256:
Currently, CI will work for PRs.

auduchinok (approved):
@DedSec256 Thanks! 🔥

PR #702, opened 2024-08-19
Remove FCS syntax tree checks from typing assists

DedSec256:


auduchinok:
@DedSec256 Is is possible to extract and merge some manageable parts of this PR? I'd like to work on the typing assists, but don't want to add conflicts or more code to rework later.

ReSharper.FSharp/src/FSharp/FSharp.Psi.Features/src/FSharpTypingAssist.fs
```diff
@@ -743,42 +742,53 @@ type FSharpTypingAssist(lifetime, dependencies) as this =
...
-                    | SynExpr.For (_, _, IdentRange range, _, _, _, _, ExprRange lastRange, _)
-
-                    // for ... in ... do {caret}
-                    | SynExpr.ForEach (_, _, _, _, PatRange range, ExprRange lastRange, _, _) when
-                            getEndOffset document lastRange = prevTokenEnd ->
-                        Some range
-
-                    | _ -> defaultTraverse expr }
+        let range =
+            let inline checkExpr (expr: IFSharpExpression) (node: ITreeNode) =
```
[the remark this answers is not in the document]
DedSec256: While I was coming up with a name for the function, I found an old bug for the case
```fsharp
for a = 0
        to 10 do  {caret}
```
its' because here
image
the range was taken not from `to ... ' expr, but `do ...`, which could never be matched with prevTokenEnd

ReSharper.FSharp/src/FSharp/FSharp.Psi.Features/src/FSharpTypingAssist.fs
```diff
@@ -736,49 +739,57 @@ type FSharpTypingAssist(lifetime, dependencies) as this =
...
-                    | SynExpr.For (_, _, IdentRange range, _, _, _, _, ExprRange lastRange, _)
+        let expr: IFSharpExpression = getNode textControl lexer.TokenStart
+        if isNull expr then false else
 
-                    // for ... in ... do {caret}
-                    | SynExpr.ForEach (_, _, _, _, PatRange range, ExprRange lastRange, _, _) when
-                            getEndOffset document lastRange = prevTokenEnd ->
-                        Some range
+        let indentOffset =
+            let inline tryGetIndentOffset (exprBeforeKeyword: IFSharpExpression) (indentAnchor: ITreeNode) =
```
auduchinok, line 753: What is the reason to inline it?

ReSharper.FSharp/src/FSharp/FSharp.Psi.Features/src/FSharpTypingAssist.fs
```diff
@@ -736,49 +739,57 @@ type FSharpTypingAssist(lifetime, dependencies) as this =
...
+                else None
 
-                    | _ -> defaultTraverse expr }
-
-        let documentCoords = document.GetCoordsByOffset(lexer.TokenStart)
-        match SyntaxTraversal.Traverse(getPosFromCoords documentCoords, parseTree, visitor) with
+            match expr with
+            // if expr then ...
+            // elif ... then {caret}
+            | :? IElifExpr as elifExpr when (elifExpr.ThenExpr :? IFromErrorExpr) ->
```
auduchinok, line 762: Should it limit to error expressions with zero length? Or do we expect it to work with erroneous expressions that could contain other ones?

What if we later decide to not convert empty FCS error expressions? I.e. the expression would be simply missing. Should we cover this case too?

ReSharper.FSharp/src/FSharp/FSharp.Psi.Features/src/FSharpTypingAssist.fs
DedSec256: In order not to allocate a closure for `tokenBeforeKeywordEnd`, used inside

ReSharper.FSharp/src/FSharp/FSharp.Psi.Features/src/FSharpTypingAssist.fs
DedSec256: Initially, when I was rewriting the code, I kept the condition with ErrorExpr, and maybe some tests didn't pass without it. The code may have changed from then.

Now I tried to remove this restriction and the tests passed!  Even more than that, it fixes the old behavior for the case 
```fsharp
if true &&
        true then {caret}5
```
which resulted in
```fsharp
if true &&
        true then
          {caret}5
```

Added tests, thanks.

ReSharper.FSharp/src/FSharp/FSharp.Psi.Features/src/FSharpTypingAssist.fs
auduchinok: And I also don't see where and how exactly it's allocated. Could you explain, please?

DedSec256: It could be viewed in Release mode in IL

image
image

In any case, if this is an obstacle on the way to merge, I will remove inline, this is not so critical.

auduchinok: Wow, I'd not expect a closure there at all. Thanks for showing this!

auduchinok: Maybe it's worth trying to find a minimal repro and reporting it to the compiler team. Or is it expected and I'm missing something?

PR #722, opened 2024-09-19
LanguageVersion: add F# 9.0

DedSec256:


ReSharper.FSharp/src/FSharp/FSharp.Common/src/ProjectModel/FSharpLanguageVersion.fs
```diff
@@ -31,7 +31,10 @@ type FSharpLanguageLevel =
     /// Nested record field copy and update/shorthand lambda
     | FSharp80 = 80
 
-    | Latest = 80
+    /// Nullness
+    | FSharp90 = 90
+
+    | Latest = 80 //TODO
```
DedSec256, line 37: If changed to 90, the tests will fail with

image

auduchinok: Thanks! We'll need to update FCS to fix it.

auduchinok (approved):
Thanks!

PR #733, opened 2024-10-09
Daemon/Stages/Patterns: add inlay type hints

DedSec256:
Currently supported: 

- Top/Local bindings & type members parameters/return types
- Foreach patterns
- Lambda parameters

Settings:

{3B3F5DCE-A9ED-4188-ABB6-25151AF82DD2}

Small demo:

example

Examples:

{FB04E510-50BA-4A42-AFCE-688ED3D851F0}

{555B8269-4C64-4012-9DFC-6CBCD12EA0F8}

{2A8F2820-1A41-4007-8091-16040B79378C}


ReSharper.FSharp/src/FSharp/FSharp.Common/src/Settings/FSharpOptions.fs
```diff
@@ -173,13 +177,23 @@ module FSharpTypeHintOptions =
...
       [<SettingsEntry(true, FSharpTypeHintOptions.hideSameLinePipe); DefaultValue>]
-      mutable HideSameLine: bool }
+      mutable HideSameLine: bool
+
+      [<SettingsEntry(PushToHintMode.Default,
+                      DescriptionResourceType = typeof<Strings>,
+                      DescriptionResourceName = nameof(Strings.FSharpTypeHints_TopLevelMembers_Description))>]
+      mutable ShowTypeHintsForTopLevelMembers: PushToHintMode
+
+      [<SettingsEntry(PushToHintMode.PushToShowHints,
```
DedSec256, line 193: Perhaps we can make a different default here, but this one seemed most appropriate to me

ReSharper.FSharp/src/FSharp/FSharp.Common/src/Settings/FSharpOptions.fs
```diff
@@ -173,13 +177,23 @@ module FSharpTypeHintOptions =
...
+                      DescriptionResourceType = typeof<Strings>,
+                      DescriptionResourceName = nameof(Strings.FSharpTypeHints_LocalBindings_Description))>]
+      mutable ShowTypeHintsForLocalBindings: PushToHintMode }
```
DedSec256, line 196: Perhaps we can make a different default here, but this one seemed most appropriate to me

ReSharper.FSharp/src/FSharp/FSharp.Psi.Daemon/src/Options/FSharpTypeHintOptionsPage.fs
```diff
@@ -0,0 +1,43 @@
...
+              "Type Hints",
+              null,
+              ParentId = nameof(FSharpInlayHintsPage),
+              NestingType = OptionPageNestingType.Child,
+              Sequence = 1.,
+              NameResourceType = typeof<Strings>,
+              NameResourceName = nameof(Strings.FSharpTypeHints_OptionsPage_Title))>]
+type FSharpTypeHintsOptionsPage(lifetime: Lifetime, optionsPageContext: OptionsPageContext,
+                                optionsSettingsSmartContext: OptionsSettingsSmartContext) as this =
+    inherit InlayHintsOptionPageBase(lifetime, optionsPageContext, optionsSettingsSmartContext, optionsPageContext.IsReSharper)
```
auduchinok, line 33: Can we move the `IsReSharper` check inside the base constructor? Though, it needs to be done outside of this PR.

ReSharper.FSharp/src/FSharp/FSharp.Psi.Daemon/src/Stages/PatternTypeHintsStage.fs
```diff
@@ -0,0 +1,257 @@
...
+
+    let isTopLevelMember (node: ITreeNode) =
+        match node with
+        | :? ITopBinding
+        | :? IMemberDeclaration
+        | :? IConstructorDeclaration -> true
+        | _ -> false
+
+    override x.VisitNode(node, context) =
+        if settings.LocalBindings = PushToHintMode.Never && isTopLevelMember node then () else
```
auduchinok, line 50: Why does it check the settings for local bindings but checks whether a node is top level?

ReSharper.FSharp/src/FSharp/FSharp.Psi.Daemon/src/Stages/PatternTypeHintsStage.fs
```diff
@@ -0,0 +1,257 @@
...
+
+        for child in node.Children() do
```
auduchinok, line 52: Does it iterate through all the children when the current mode should only show the hints for the top level nodes? Or do we need to create the hints anyway?

ReSharper.FSharp/src/FSharp/FSharp.Psi.Daemon/src/Stages/PatternTypeHintsStage.fs
```diff
@@ -0,0 +1,257 @@
...
+
+        let symbolUse =
+            match decl with
+            | :? IBinding as binding ->
+                let refPat = binding.HeadPattern.As<IReferencePat>()
+                if isNull refPat then Unchecked.defaultof<_> else
+                refPat.GetFcsSymbolUse()
+            | :? IMemberDeclaration as memberDeclaration ->
+                memberDeclaration.GetFcsSymbolUse()
+            | _ -> Unchecked.defaultof<_>
```
auduchinok, line 136: Would getting it via `IFSharpDeclaration` work better?

ReSharper.FSharp/src/FSharp/FSharp.Psi.Daemon/src/Stages/PatternTypeHintsStage.fs
```diff
@@ -0,0 +1,257 @@
...
+
+        if isNull symbolUse then ValueNone else
+        let symbol = symbolUse.Symbol.As<FSharpMemberOrFunctionOrValue>()
```
auduchinok, line 139: Could you please add empty lines before such `else` branches?

ReSharper.FSharp/src/FSharp/FSharp.Psi.Daemon/src/Stages/PatternTypeHintsStage.fs
```diff
@@ -0,0 +1,257 @@
...
+
+        | :? IAutoPropertyDeclaration as decl ->
+            let symbolUse = decl.GetFcsSymbolUse()
+            if isNull symbolUse then ValueNone else
+            let symbol = symbolUse.Symbol.As<FSharpMemberOrFunctionOrValue>()
+            if isNull symbol then ValueNone else
+
+            let fcsType = symbol.ReturnParameter.Type
+            let range = decl.NameIdentifier.GetNavigationRange().EndOffsetRange()
+            let displayContext = symbolUse.DisplayContext.WithShortTypeNames(true)
```
auduchinok, line 186: Parts of this code seem to be repeated here and there. Could you try to extract it and reuse?

ReSharper.FSharp/src/FSharp/FSharp.Psi.Daemon/src/Stages/PatternTypeHintsStage.fs
```diff
@@ -0,0 +1,257 @@
...
+        adornNodes localNodes settings.LocalBindings
+
+        highlightingConsumer.CollectHighlightings()
+
+    override x.Execute(committer) =
+        let consumer = { TopLevelNodes = List(); LocalNodes = List() }
+        fsFile.Accept(MembersVisitor(settings), consumer)
+
+        let topLevelNodes = consumer.TopLevelNodes |> Array.ofSeq
+        let localNodes = consumer.LocalNodes |> Array.ofSeq
```
auduchinok, line 214: Could you explain why the arrays are needed, please?

ReSharper.FSharp/src/FSharp/FSharp.Psi.Services/src/Util/TypeAnnotationsUtil.fs
```diff
@@ -0,0 +1,90 @@
...
+    |> Seq.fold visitPattern acc
+
+let collectTypeHintAnchorsForBinding (binding: IBinding) =
+    let acc: ITreeNode list = []
+
+    if isNull binding || isNull binding.EqualsToken then acc else
+
+    let acc =
+        if isNotNull binding.ReturnTypeInfo then acc
+        elif binding.HasParameters then [(binding: ITreeNode)]
```
auduchinok, line 51: Are parens needed here?

ReSharper.FSharp/src/FSharp/FSharp.Psi/src/FSharp.psi
```diff
@@ -1457,6 +1457,7 @@ autoPropertyDeclaration options { stubBase="FSharpProperTypeMemberDeclarationBas
   memberKeyword{MEMBER, MemberKeyword}
   VAL<VAL, ValKeyword>
   IDENTIFIER<IDENTIFIER, Identifier>
+  typeUsage<TYPE, TypeUsage>?
```
auduchinok, line 1460: Isn't the type usage separated from the name with a `:`?

ReSharper.FSharp/src/FSharp/FSharp.Psi/src/Impl/Tree/PrimaryConstructorDeclaration.cs
```diff
@@ -13,6 +13,8 @@ public override TreeTextRange GetNameRange() =>
     protected override IDeclaredElement CreateDeclaredElement() =>
       new FSharpPrimaryConstructor(this);
 
+    public ITokenNode EqualsToken => null;
```
auduchinok, line 16: What do you think about moving the following here, from `collectTypeHintAnchorsForConstructor`?
```fsharp
let owner = FSharpTypeDeclarationNavigator.GetByPrimaryConstructorDeclaration(ctor)
isNotNull owner && isNotNull owner.EqualsToken
```

We could then have a doc comment saying something like ``Returns associated `=` token, possibly from containing type declaration``.

auduchinok:
@DedSec256 Very nice work! What do you think about moving `git` authors to the right of declarations, so we could prevent the interlines in the editor?

ReSharper.FSharp/src/FSharp/FSharp.Psi.Daemon/src/Stages/PatternTypeHintsStage.fs
DedSec256: The idea is not to visit children inside the top-level node if hints are not needed inside anyway.

ReSharper.FSharp/src/FSharp/FSharp.Psi.Daemon/src/Stages/PatternTypeHintsStage.fs
DedSec256: If we want to show hints only for the top level, then the check above will not allow us to go inside the top binding. We create hints for children if the Always/PushToHint mode is enabled for local bindings.

ReSharper.FSharp/src/FSharp/FSharp.Psi.Daemon/src/Stages/PatternTypeHintsStage.fs
DedSec256: Because of `Array.partition` function. Made by analogy with

https://github.com/JetBrains/resharper-fsharp/blob/f6475da23004a3eb8f2487e8a161e11ead450c26/ReSharper.FSharp/src/FSharp/FSharp.Psi.Daemon/src/Stages/PipeChainTypeHintStage.fs#L150-L164

Perhaps there is a more efficient way, but I didn’t immediately come up with a more elegant solution

ReSharper.FSharp/src/FSharp/FSharp.Psi.Services/src/Util/TypeAnnotationsUtil.fs
DedSec256: Yeah 

{08508C3B-125C-4CC3-B53C-A395ADD78022}

{651D424A-581C-4B9F-8593-01C0C2B2737A}


it is also possible to write `binding :> ITreeNode`

ReSharper.FSharp/src/FSharp/FSharp.Psi/src/FSharp.psi
DedSec256: Sorry, this change should not be presented in **this PR**. I will remove it.

ReSharper.FSharp/src/FSharp/FSharp.Psi.Daemon/src/Stages/PatternTypeHintsStage.fs
DedSec256: Oh, right now bindings are not an IFSharpDeclaration :(

ReSharper.FSharp/src/FSharp/FSharp.Psi.Daemon/src/Stages/PatternTypeHintsStage.fs
DedSec256: It's good that you noticed. In general, this entire piece of code should be deleted, since we don't show hints for auto-properties (just like we didn't show Code Vision)

ReSharper.FSharp/src/FSharp/FSharp.Psi.Services/src/Util/TypeAnnotationsUtil.fs
auduchinok: Wow, OK. 😞

ReSharper.FSharp/src/FSharp/FSharp.Psi.Daemon/FSharp.Psi.Daemon.fsproj
```diff
@@ -13,8 +13,6 @@
     <TargetFramework>net472</TargetFramework>
   </PropertyGroup>
   <ItemGroup>
-    <Compile Include="PsiFeaturesNamespaceZoneMarker.fs" />
-    <Compile Include="PsiNamespaceZoneMarker.fs" />
```
auduchinok, line 17: Are these changes needed in this PR?

ReSharper.FSharp/src/FSharp/FSharp.Psi.Daemon/src/Options/FSharpTypeHintOptionsPage.fs
auduchinok: @DedSec256 Please move the check inside the base constructor after this is merged to the monorepo.

ReSharper.FSharp/src/FSharp/FSharp.Psi.Daemon/src/Stages/PatternTypeHintsStage.fs
```diff
@@ -0,0 +1,250 @@
...
+        context.LocalNodes.AddRange(result)
+
+        x.VisitNode(forEachExpr, context)
+
+type private PatternsHighlightingProcess(logger: ILogger, fsFile, settingsStore: IContextBoundSettingsStore,
+                                           daemonProcess: IDaemonProcess, settings) =
+    inherit FSharpDaemonStageProcessBase(fsFile, daemonProcess)
+
+    let createTypeHintHighlighting (fcsType: FSharpType) (displayContext: FSharpDisplayContext) range pushToHintMode =
+        let displayContext = displayContext.WithShortTypeNames(true)
```
auduchinok, line 107: I wonder if we could reuse the display context between symbols. We use short names anyway, so the context is not needed (much?), as the main thing it brings here is what namespaces are open.

ReSharper.FSharp/src/FSharp/FSharp.Psi.Daemon/src/Stages/PatternTypeHintsStage.fs
```diff
@@ -0,0 +1,250 @@
...
+        TypeHintHighlighting(fcsType.Format(displayContext), range, pushToHintMode)
+
+    let getReturnTypeHint (decl: IParameterOwnerMemberDeclaration) pushToHintMode =
+        match decl with
+        | :? IConstructorDeclaration
+        | :? IAccessorDeclaration -> ValueNone
+        | _ ->
+
+        let equalsToken = decl.EqualsToken
+        let patterns = decl.ParameterPatterns
```
auduchinok, line 117: Do we really need to enumerate all the patterns and store them in a collection? Can we just check that parameter declaration sequence is not empty?

ReSharper.FSharp/src/FSharp/FSharp.Psi.Daemon/src/Stages/PatternTypeHintsStage.fs
```diff
@@ -0,0 +1,250 @@
...
+                memberDeclaration.NameIdentifier.GetDocumentRange().EndOffsetRange()
+
+            | _ -> equalsToken.GetDocumentRange().StartOffsetRange()
+
+        let symbolUse =
+            match decl with
+            | :? IBinding as binding ->
+                let refPat = binding.HeadPattern.As<IReferencePat>()
+                if isNull refPat then Unchecked.defaultof<_> else
+                refPat.GetFcsSymbolUse()
```
auduchinok, line 134: Nit: a `match` expression could probably look cleaner.

ReSharper.FSharp/src/FSharp/FSharp.Psi.Daemon/src/Stages/PatternTypeHintsStage.fs
```diff
@@ -0,0 +1,250 @@
...
+        | :? IParametersOwnerPat as pattern ->
+            let asPat = AsPatNavigator.GetByLeftPattern(pattern.IgnoreParentParens())
+            if isNull asPat then ValueNone else
+
+            let reference = pattern.Reference
+            if isNull reference then ValueNone else
+
+            let symbol = reference.GetFcsSymbol().As<FSharpActivePatternCase>()
+            if isNull symbol then ValueNone
+            else getHintForPattern (asPat.RightPattern.IgnoreInnerParens()) pushToHintMode
```
auduchinok, line 158: Same thing about a possible `match` expression.

ReSharper.FSharp/src/FSharp/FSharp.Psi.Daemon/src/Stages/PatternTypeHintsStage.fs
auduchinok: I'm fine with merging it as is for now, but we should really think about rewriting this and that pieces of code.

auduchinok (approved):

ReSharper.FSharp/src/FSharp/FSharp.Psi.Daemon/FSharp.Psi.Daemon.fsproj
DedSec256: I touched fsproj around this place and decided to bring the position of neighboring files to a more expected state. We can also do this in a separate PR.

ReSharper.FSharp/src/FSharp/FSharp.Psi.Daemon/src/Stages/PatternTypeHintsStage.fs
DedSec256: Maybe we could just use a default context with short names?

auduchinok: If there's one, and it works with it, then, for sure, let's do it.

PR #756, opened 2024-10-31
Inlay hints/PatternTypeHintsStage: support other patterns

DedSec256:
{BEB256F4-5DD9-481A-8B66-8929F26EA3BB}

- [ ] FCS update

### DIFF
--- a/ReSharper.FSharp/src/FSharp/FSharp.Common/src/Settings/FSharpOptions.fs
+++ b/ReSharper.FSharp/src/FSharp/FSharp.Common/src/Settings/FSharpOptions.fs
@@ -196,7 +196,7 @@ type FSharpTypeHintOptions =
       [<SettingsEntry(PushToHintMode.PushToShowHints,
                       DescriptionResourceType = typeof<Strings>,
                       DescriptionResourceName = nameof(Strings.FSharpTypeHints_OtherPatterns_Description))>]
-      mutable ShowTypeHintsForOtherPatterns: PushToHintMode }
+      mutable ShowForOtherPatterns: PushToHintMode }
 
 
 [<OptionsPage("FSharpOptionsPage", "F#", typeof<ProjectModelThemedIcons.Fsharp>)>]
--- a/ReSharper.FSharp/src/FSharp/FSharp.Common/src/Settings/FSharpOptions.fs
+++ b/ReSharper.FSharp/src/FSharp/FSharp.Common/src/Settings/FSharpOptions.fs
@@ -191,7 +191,12 @@ type FSharpTypeHintOptions =
       [<SettingsEntry(PushToHintMode.PushToShowHints,
                       DescriptionResourceType = typeof<Strings>,
                       DescriptionResourceName = nameof(Strings.FSharpTypeHints_LocalBindings_Description))>]
-      mutable ShowTypeHintsForLocalBindings: PushToHintMode }
+      mutable ShowTypeHintsForLocalBindings: PushToHintMode
+
+      [<SettingsEntry(PushToHintMode.PushToShowHints,
+                      DescriptionResourceType = typeof<Strings>,
+                      DescriptionResourceName = nameof(Strings.FSharpTypeHints_OtherPatterns_Description))>]
+      mutable ShowTypeHintsForOtherPatterns: PushToHintMode }
 
 
 [<OptionsPage("FSharpOptionsPage", "F#", typeof<ProjectModelThemedIcons.Fsharp>)>]
--- a/ReSharper.FSharp/src/FSharp/FSharp.Psi.Daemon/src/Stages/PatternTypeHintsStage.fs
+++ b/ReSharper.FSharp/src/FSharp/FSharp.Psi.Daemon/src/Stages/PatternTypeHintsStage.fs
@@ -22,26 +22,26 @@ open JetBrains.ReSharper.Plugins.FSharp.Psi.Services.Util.TypeAnnotationsUtil
 open JetBrains.TextControl.DocumentMarkup.Adornments.IntraTextAdornments
 
 type private NodesRequiringHints =
-    { TopLevelNodes: VisibilityConsumer<ITreeNode>
-      LocalNodes: VisibilityConsumer<ITreeNode>
-      OtherPatterns: VisibilityConsumer<ITreeNode> } with
+    { TopLevelMembers: VisibilityConsumer<ITreeNode>
+      LocalBindings: VisibilityConsumer<ITreeNode>
+      OtherPatterns: VisibilityConsumer<ITreeNode> }
 
     member x.HasVisibleItems =
-        x.TopLevelNodes.HasVisibleItems ||
-        x.LocalNodes.HasVisibleItems ||
+        x.TopLevelMembers.HasVisibleItems ||
+        x.LocalBindings.HasVisibleItems ||
         x.OtherPatterns.HasVisibleItems
 
 type private FSharpTypeHintSettings =
     { TopLevelMembers: PushToHintMode
       LocalBindings: PushToHintMode
-      OtherPatterns: PushToHintMode } with
+      OtherPatterns: PushToHintMode }
 
     static member Create(settingsStore: IContextBoundSettingsStore) =
         { TopLevelMembers = settingsStore.GetValue(fun (key: FSharpTypeHintOptions) -> key.ShowTypeHintsForTopLevelMembers)
                                          .EnsureInlayHintsDefault(settingsStore)
           LocalBindings = settingsStore.GetValue(fun (key: FSharpTypeHintOptions) -> key.ShowTypeHintsForLocalBindings)
                                        .EnsureInlayHintsDefault(settingsStore)
-          OtherPatterns = settingsStore.GetValue(fun (key: FSharpTypeHintOptions) -> key.ShowTypeHintsForOtherPatterns)
+          OtherPatterns = settingsStore.GetValue(fun (key: FSharpTypeHintOptions) -> key.ShowForOtherPatterns)
                                        .EnsureInlayHintsDefault(settingsStore)}
 
     member x.IsDisabled =
@@ -52,7 +52,7 @@ type private FSharpTypeHintSettings =
 
 type private MembersVisitor(settings) =
     inherit TreeNodeVisitor<NodesRequiringHints>()
-    let disabledForTopBindings = settings.TopLevelMembers = PushToHintMode.Never
+    let disabledForTopMembers = settings.TopLevelMembers = PushToHintMode.Never
     let disabledForLocalBindings = settings.LocalBindings = PushToHintMode.Never
     let disabledForOtherPatterns = settings.OtherPatterns = PushToHintMode.Never
 
@@ -73,7 +73,7 @@ type private MembersVisitor(settings) =
         if disabledForLocalBindings && disabledForOtherPatterns && isTopLevelMember node then () else
 
         for child in node.Children() do
-            if disabledForTopBindings && isTopLevelMember child || disabledForLocalBindings && isLocalBinding child
+            if disabledForTopMembers && isTopLevelMember child || disabledForLocalBindings && isLocalBinding child
                 then x.VisitNode(child, context) else
 
             match child with
@@ -82,37 +82,37 @@ type private MembersVisitor(settings) =
 
     override x.VisitLambdaExpr(lambda, context) =
         let result = collectTypeHintAnchorsForLambda lambda
-        context.LocalNodes.AddRange(result)
+        context.LocalBindings.AddRange(result)
 
         x.VisitNode(lambda, context)
 
     override x.VisitLocalBinding(binding, context) =
         let result = collectTypeHintAnchorsForBinding binding
-        context.LocalNodes.AddRange(result)
+        context.LocalBindings.AddRange(result)
 
         x.VisitNode(binding, context)
 
     override x.VisitTopBinding(binding, context) =
         let result = collectTypeHintAnchorsForBinding binding
-        context.TopLevelNodes.AddRange(result)
+        context.TopLevelMembers.AddRange(result)
 
         x.VisitNode(binding, context)
 
     override x.VisitMemberDeclaration(memberDecl, context) =
         let result = collectTypeHintsAnchorsForMember memberDecl
-        context.TopLevelNodes.AddRange(result)
+        context.TopLevelMembers.AddRange(result)
 
         x.VisitNode(memberDecl, context)
 
     override x.VisitPrimaryConstructorDeclaration(constructor, context) =
         let result = collectTypeHintAnchorsForConstructor constructor
-        context.TopLevelNodes.AddRange(result)
+        context.TopLevelMembers.AddRange(result)
 
         x.VisitNode(constructor, context)
 
     override x.VisitSecondaryConstructorDeclaration(constructor, context) =
         let result = collectTypeHintAnchorsForConstructor constructor
-        context.TopLevelNodes.AddRange(result)
+        context.TopLevelMembers.AddRange(result)
 
         x.VisitNode(constructor, context)
 
@@ -228,8 +228,8 @@ type private PatternsHighlightingProcess(fsFile, settingsStore: IContextBoundSet
         | _ -> ValueNone
 
     let adornNodes
-        (topLevelNodes: ITreeNode ICollection)
-        (localNodes: ITreeNode ICollection)
+        (topLevelMembers: ITreeNode ICollection)
+        (localBindings: ITreeNode ICollection)
         (otherPatterns: ITreeNode ICollection) =
         let highlightingConsumer = FilteringHighlightingConsumer(daemonProcess.SourceFile, fsFile, settingsStore)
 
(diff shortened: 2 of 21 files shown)


auduchinok:
@DedSec256 Do you think it could make sense to introduce another option, so hints for bindings/parameter declarations could be controlled separately? With both currently defaulting to 'push-to-hint' it doesn't make much difference, but I'm also trying to imagine if somebody could want to make the declaration hints be always shown while keep the 'match' ones in push-to-hint mode.

ReSharper.FSharp/test/data/features/daemon/typeHints/Signatures - Top level 01.fs.gold
```diff
@@ -41,34 +41,34 @@ let g||(15) = fun x -> ()
 
 type A(x||(16)) =
     do
-        let x||(17) = 3
```
DedSec256, line 44: Fixed the case when hints showed for some local bindings even if it was disabled

DedSec256:
Perhaps, we can write  `Match clause patterns`

image

auduchinok:
> Perhaps, we can name it  Match clause patterns
"Clause" sounds like a rather technical term to me. I've never heard it from non-compiler people. Something like "patterns" or "branches" may sound much more user-friendly here, I think?

ReSharper.FSharp/test/data/features/daemon/typeHints/Signatures - Locals 01.fs.gold
```diff
@@ -130,6 +130,21 @@ let (|String|) (x: MyStruct) : string = String(x.myString)
 
 let f1 (x & Int(i) & String(s)) = ()
 
+
+let _ = function x -> ()
```
auduchinok, line 134: What do you think about splitting this test files into smaller ones, so it could be easier to think about unrelated cases separately?

ReSharper.FSharp/test/data/features/daemon/typeHints/Signatures - Locals 01.fs.gold
```diff
@@ -130,6 +130,21 @@ let (|String|) (x: MyStruct) : string = String(x.myString)
...
+
+match [Some 5] with
+| [] -> ()
+| [x] -> ()
```
auduchinok, line 138: Are we going to add an annotation for `x` here? Assuming the matched expression type isn't obvious, a hint would help here a lot.

ReSharper.FSharp/test/data/features/daemon/typeHints/Signatures - Match clauses 01.fs.gold
```diff
@@ -0,0 +1,153 @@
...
+
+    member _.M(x: int) (y: int, z: int) : int =
+        let res: int = x + y + z
+        res
+
+    interface IDisposable with
+        member _.Dispose(): unit = ()
+
+let delimiter (delim1, delim2, value) =
+    { new IFormattable with
```
auduchinok, line 106: Is it really a part of `match`-related test?

If the idea here is  to test how different settings work, can these tests be simplified to only include minimal code needed for such tests?

ReSharper.FSharp/test/data/features/daemon/typeHints/Signatures - Locals 01.fs.gold
DedSec256: @auduchinok, how would you expect to see hints for 

```fsharp
| [x; y] -> ...
```
?

ReSharper.FSharp/test/data/features/daemon/typeHints/Signatures - Match clauses 01.fs.gold
DedSec256: I'd like to suggest we refactor these tests in a separate pull request.

auduchinok: Yes, sounds good. Maybe we should merge the refactor first then.

ReSharper.FSharp/test/data/features/daemon/typeHints/Signatures - Locals 01.fs.gold
auduchinok: That's a good question. Also consider this:

```fsharp
| x :: _ -> ...
| x :: y :: _ -> ...

| [| x |] -> ...
| [| x; y |[ -> ...

| [] -> ...
| [||] -> ...
```

Maybe annotating the first item is fine. Maybe annotating the collection is better.

DedSec256: I can offer a universal heuristic: if there is "non-obvious" patterns count  != 1, then annotate the collection, otherwise annotate a single pattern

image

DedSec256: I can suggest the following heuristic: if the number of patterns that require direct annotation is 1, then annotate the pattern, otherwise annotate the collection

image

DedSec256: Some real-world examples

image
image
image
image

DedSec256: perhaps in the case 
```fsharp
| head :: tail - > ...
```
it makes sense to give priority to the head pattern

auduchinok: > I can suggest the following heuristic: if the number of top-level patterns in the collection that require direct annotation is 1, then annotate the pattern, otherwise annotate the collection

I would vote for consistency instead of heuristics.
Consider the following:
```fsharp
match f x with
| [] -> ...
| [x] -> ...
| [x; y] -> ...
```
Why do we want these three simple patterns to have different hint placements? Are there any realistic common cases where `[x]` is not going to be placed along the other patterns in the same match?
Are there any actual benefits of `[x]` having a different style?

DedSec256: Then it might look like this 

image

There are a couple more questions, for example
- Should we always annotate `[]`, because it doesn't contain elements that you can work with, which makes it similar to `_` pattern. It might make sense to show hints only for the `[] as x`
- Should we show an annotation for a collection if there is an explicitly annotated pattern inside
`[x; y: int option]`

> Are there any realistic common cases where [x] is not going to be placed along the other patterns in the same match?

In a match, they can follow each other, but the bodies of the corresponding branches can take enough lines so that the clauses are not located nearby

auduchinok: > In a match, they can follow each other, but the bodies of the corresponding branches can take enough lines so that the clauses are not located nearby

In what way does it change something? What are the benefits of having separate styles?

DedSec256: I don't mind showing hints uniformly (I've attached an example above of how it might look).

Initially, I was guided by the idea to show hints near points of interest. If there are several of them and they will duplicate the same hints, then show a hint in one place, for the collection.

DedSec256: @auduchinok, what do you think about https://github.com/JetBrains/resharper-fsharp/pull/756#discussion_r1877182745

auduchinok: @DedSec256 It looks good.

ReSharper.FSharp/src/FSharp/FSharp.Common/src/Settings/FSharpOptions.fs
```diff
@@ -194,7 +194,12 @@ type FSharpTypeHintOptions =
       [<SettingsEntry(PushToHintMode.PushToShowHints,
                       DescriptionResourceType = typeof<Strings>,
                       DescriptionResourceName = nameof(Strings.FSharpTypeHints_LocalBindings_Description))>]
-      mutable ShowTypeHintsForLocalBindings: PushToHintMode }
+      mutable ShowTypeHintsForLocalBindings: PushToHintMode
+
+      [<SettingsEntry(PushToHintMode.PushToShowHints,
+                      DescriptionResourceType = typeof<Strings>,
+                      DescriptionResourceName = nameof(Strings.FSharpTypeHints_MatchPatterns_Description))>]
+      mutable ShowTypeHintsForMatchPatterns: PushToHintMode }
```
auduchinok, line 202: What's the idea behind having separate settings for patterns in local bindings and match clauses?

ReSharper.FSharp/src/FSharp/FSharp.Psi.Daemon/src/Stages/PatternTypeHintsStage.fs
```diff
@@ -214,22 +251,26 @@ type private PatternsHighlightingProcess(fsFile, settingsStore: IContextBoundSet
         // Intersect them to ensure commit doesn't throw
         let documentRange = daemonProcess.Document.GetDocumentRange()
         let visibleRange = daemonProcess.VisibleRange.Intersect(&documentRange)
-        let consumer = { TopLevelNodes = VisibilityConsumer(visibleRange, _.GetNavigationRange())
-                         LocalNodes = VisibilityConsumer(visibleRange, _.GetNavigationRange()) }
+        let consumer = {
+                         TopLevelNodes = VisibilityConsumer(visibleRange, _.GetNavigationRange())
```
auduchinok, line 255: Please reformat.

ReSharper.FSharp/src/FSharp/FSharp.Psi.Daemon/src/Stages/PatternTypeHintsStage.fs
```diff
@@ -181,7 +196,25 @@ type private PatternsHighlightingProcess(fsFile, settingsStore: IContextBoundSet
             createTypeHintHighlighting fcsType defaultDisplayContext range pushToHintMode actionsProvider false
             |> ValueSome
 
-        | _ -> ValueNone
+        | :? IParametersOwnerPat as pattern ->
```
auduchinok, line 199: Any reason for this change?

ReSharper.FSharp/test/data/features/daemon/typeHints/Patterns 02 - Match clauses.fs.gold
```diff
@@ -0,0 +1,82 @@
...
+| x :: (y :: _||(19)) -> ()
+| x :: y :: []||(20) -> ()
+| x :: [y]||(21) -> ()
+| Some(x||(22)) :: tail||(23) -> () 
+| [Some(x||(24))]||(25)
+| [_; Some(x)] when let x||(26) = 5 in true ->
+    let y||(27) = 5 in ()
+
+match [[5]] with
+| [[]||(28)]||(29) -> ()
```
auduchinok, line 35: Do you think it makes sense to do it recursively here?

ReSharper.FSharp/test/data/features/daemon/typeHints/Patterns 02 - Match clauses.fs.gold
```diff
@@ -0,0 +1,82 @@
...
+| _ :: (tail: int option list) -> ()
+| x: int option :: tail||(11) -> ()
+| x: int option :: _||(12) -> ()
+| x :: (tail: int option list) -> ()
+| ((x :: tail): int option list) -> ()
+| _ :: _||(13) -> ()
+| x :: _||(14) -> ()
+| _ :: x :: _||(15) -> ()
+| x :: _ :: _||(16) -> ()
+| x :: Some y||(17) :: _||(18) -> ()
```
auduchinok, line 25: Do you think it makes sense to do it recursively here?

ReSharper.FSharp/test/data/features/daemon/typeHints/Settings 01 - Show all.fs.gold
```diff
@@ -6,37 +6,43 @@ let f x||(0) ||(1)=
...
-type A(x||(11)) =
+match 5 with
+| x||(12) when let y||(13) = 5 in true ->
+    let z||(14) = 3
+    ()
+
+type A(x||(15)) =
     do
-        let x||(12) = 3
+        let x||(16) = 3
```
auduchinok, line 22: @DedSec256 This seems unrelated to match clauses. Could you split these test cases, please?

ReSharper.FSharp/src/FSharp/FSharp.Common/src/Settings/FSharpOptions.fs
DedSec256: I think we discussed this question here https://github.com/JetBrains/resharper-fsharp/pull/756#issuecomment-2462035720, or was it about something else?

ReSharper.FSharp/test/data/features/daemon/typeHints/Patterns 02 - Match clauses.fs.gold
DedSec256: I think for such cases I would like to see annotations for inner patterns as well

image

auduchinok: All the inferred types are going to be shown in the outer hint anyway, though. I'm not sure if it's helpful to show it for non-generic fields types. While it might help somewhat to see the types next to the patterns, I think about how to prevent such hints from becoming too noisy.

ReSharper.FSharp/test/data/features/daemon/typeHints/Settings 01 - Show all.fs.gold
DedSec256: This is a special test case where all hints are enabled and contains the minimum set of each hint category. Previously I split tests into smaller ones here
https://github.com/JetBrains/resharper-fsharp/commit/313392c557b7e5c7ef2774676a40de77c9196e05

ReSharper.FSharp/src/FSharp/FSharp.Psi.Daemon/src/Stages/PatternTypeHintsStage.fs
```diff
@@ -21,28 +22,39 @@ open JetBrains.ReSharper.Plugins.FSharp.Psi.Services.Util.TypeAnnotationsUtil
 open JetBrains.TextControl.DocumentMarkup.Adornments.IntraTextAdornments
 
 type private NodesRequiringHints =
-    { TopLevelNodes: VisibilityConsumer<ITreeNode>; LocalNodes: VisibilityConsumer<ITreeNode> } with
+    { TopLevelNodes: VisibilityConsumer<ITreeNode>
+      LocalNodes: VisibilityConsumer<ITreeNode>
+      OtherPatterns: VisibilityConsumer<ITreeNode> } with
```
auduchinok, line 27: This naming seems inconsistent a bit. Shouldn't the nodes from `OtherPatterns` be the subset of `LocalNodes`, according to this naming? Or are `LocalNodes` about binding patterns specifically?

ReSharper.FSharp/src/FSharp/FSharp.Psi.Daemon/src/Stages/PatternTypeHintsStage.fs
```diff
@@ -21,28 +22,39 @@ open JetBrains.ReSharper.Plugins.FSharp.Psi.Services.Util.TypeAnnotationsUtil
...
         x.TopLevelNodes.HasVisibleItems ||
-        x.LocalNodes.HasVisibleItems
+        x.LocalNodes.HasVisibleItems ||
+        x.OtherPatterns.HasVisibleItems
 
 type private FSharpTypeHintSettings =
-    { TopLevelMembers: PushToHintMode; LocalBindings: PushToHintMode } with
+    { TopLevelMembers: PushToHintMode
+      LocalBindings: PushToHintMode
+      OtherPatterns: PushToHintMode } with
```
auduchinok, line 37: Is there a reason `with` keyword is required here? Should we have an analyzer/QF for removing it?

ReSharper.FSharp/src/FSharp/FSharp.Psi.Daemon/src/Stages/PatternTypeHintsStage.fs
```diff
@@ -51,12 +63,18 @@ type private MembersVisitor(settings) =
         | :? IConstructorDeclaration -> true
         | _ -> false
 
+    let isLocalBinding (node: ITreeNode) =
+        match node with
+        | :? ILocalBinding
+        | :? ILambdaExpr -> true
```
auduchinok, line 69: This is a corner case, but in some other cases we do consider these patterns to be somewhat top-level:
```fsharp
module Module

let f1 = fun x -> x + 1
let f1 x = fun y -> x + y
```

ReSharper.FSharp/src/FSharp/FSharp.Common/src/Settings/FSharpOptions.fs
```diff
@@ -191,7 +191,12 @@ type FSharpTypeHintOptions =
       [<SettingsEntry(PushToHintMode.PushToShowHints,
                       DescriptionResourceType = typeof<Strings>,
                       DescriptionResourceName = nameof(Strings.FSharpTypeHints_LocalBindings_Description))>]
-      mutable ShowTypeHintsForLocalBindings: PushToHintMode }
+      mutable ShowTypeHintsForLocalBindings: PushToHintMode
+
+      [<SettingsEntry(PushToHintMode.PushToShowHints,
+                      DescriptionResourceType = typeof<Strings>,
+                      DescriptionResourceName = nameof(Strings.FSharpTypeHints_OtherPatterns_Description))>]
+      mutable ShowTypeHintsForOtherPatterns: PushToHintMode }
```
auduchinok, line 199: Nit-picking: `ShowTypeHintsForOtherPatterns` sounds a bit tautological in the context of `FSharpTypeHintOptions`. Could it maybe be something like `ShowOtherPatterns`/`OtherPatternsMode`? Ideally, this should be inline with the other settings, so I'm not sure if it's feasible to change it without breaking existing settings or unneeded migrations.

auduchinok (approved):
@DedSec256 Very good work, thanks! I left some comments, feels free to take these into account before merging

ReSharper.FSharp/src/FSharp/FSharp.Psi.Daemon/src/Stages/PatternTypeHintsStage.fs
DedSec256: What do you think about doing the same logic for this case here in a separate PR?

auduchinok: Sure, as you wish 🙂

PR #858, opened 2025-05-07
FSharpLanguageVersion: invalidate on toolset change

DedSec256:


ReSharper.FSharp/src/FSharp/FSharp.Psi.Daemon/src/Stages/FSharpFileStructureExplorer.fs
```diff
@@ -49,4 +50,5 @@ type FSharpFileStructureExplorer() =
             let fsFile = file.As<IFSharpFile>()
             if isNull fsFile then null else
 
+            PsiFileCachedDataUtil.InvalidateAllData(fsFile)
```
DedSec256, line 53: Made by analogy with C#, clears the cached LanguageVersion in the file

auduchinok (approved):
Nice!

PR #874, opened 2025-06-03
Specify Types: auto-import types 

DedSec256:
- [x] Use a qualified name if has conflicts (obsolete)


ReSharper.FSharp/src/FSharp/FSharp.Psi.Intentions/src/Intentions/FunctionAnnotationAction.fs
```diff
@@ -35,6 +36,52 @@ module SpecifyTypes =
...
         static member ReturnTypeOnly = { CanSpecifyParameterTypes = false; CanSpecifyReturnType = true }
 
+    let rec bind (context: ITreeNode) (fcsType: FSharpType) (typeUsage: ITypeUsage) =
+        if fcsType.IsGenericParameter then () else
+
+        match typeUsage with
+        | :? IAnonRecordTypeUsage as typeUsage ->
+            typeUsage.Fields |> Seq.iteri (fun i field ->
+                let typeUsage = field.TypeUsage
+                bind context fcsType.GenericArguments[i] typeUsage)
```
auduchinok: This is really unexpected! I wouldn't call the record field types type arguments here. On the other hand, there's no entity that defines the type to look at the fields.  I'd probably expect the types to be in the `AnonRecordTypeDetails` instead. 🙂 

Anyway, it's good that these types are accessible here, but it looks a bit like an exposed implementation detail to me. Let's add some more tests here, please?

ReSharper.FSharp/src/FSharp/FSharp.Psi.Intentions/src/Intentions/FunctionAnnotationAction.fs
```diff
@@ -35,6 +36,52 @@ module SpecifyTypes =
...
+            bind context fcsType typeUsage.InnerTypeUsage
+
+        | :? IFunctionTypeUsage as typeUsage ->
+            let typeUsage1 = typeUsage.ArgumentTypeUsage
+            let argType = fcsType.GenericArguments[0]
+            bind context argType typeUsage1
+
+            let typeUsage2 = typeUsage.ReturnTypeUsage
+            let returnType = fcsType.GenericArguments[1]
+            bind context returnType typeUsage2
```
auduchinok: Let's also have more tests for recursive tree structure like this, so it'd be safer to change how things are represented if needed.

ReSharper.FSharp/src/FSharp/FSharp.Psi.Intentions/src/Intentions/FunctionAnnotationAction.fs
```diff
@@ -35,6 +36,52 @@ module SpecifyTypes =
...
+        | :? IArrayTypeUsage as typeUsage ->
+            let typeUsage = typeUsage.TypeUsage
+            let fcsType = fcsType.GenericArguments[0]
+            bind context fcsType typeUsage
+
+        | :? INamedTypeUsage as typeUsage ->
+            let typeReference = typeUsage.ReferenceName
+            let reference = typeReference.Reference
+            let symbol = fcsType.TypeDefinition
+            FSharpBindUtil.bindFcsSymbolToReference context reference symbol "SpecifyTypesAction"
```
auduchinok: Side note: may be we should assert that this happens either during a read operation or that there are pinned check results.

ReSharper.FSharp/src/FSharp/FSharp.Psi.Intentions/src/Intentions/FunctionAnnotationAction.fs
```diff
@@ -35,6 +36,52 @@ module SpecifyTypes =
...
+            let typeUsage2 = typeUsage.ReturnTypeUsage
+            let returnType = fcsType.GenericArguments[1]
+            bind context returnType typeUsage2
+
+        | :? IArrayTypeUsage as typeUsage ->
+            let typeUsage = typeUsage.TypeUsage
+            let fcsType = fcsType.GenericArguments[0]
+            bind context fcsType typeUsage
+
+        | :? INamedTypeUsage as typeUsage ->
```
auduchinok: I'd personally probably either move this case to top as the most important one, or would sort the cases alphabetically. Feel free to ignore this 🙂

ReSharper.FSharp/src/FSharp/FSharp.Psi.Services/src/Util/FSharpResolveUtil.fs
```diff
@@ -28,15 +28,14 @@ let private resolvesToAssociatedModule (declaredElement: IDeclaredElement) (unqu
     declaredElement.Equals(typeElement)
 
 let private resolvesTo (declaredElement: IDeclaredElement) (reference: FSharpSymbolReference) qualified resolveExpr opName =
-    match reference.ResolveWithFcs(opName, resolveExpr, qualified) with
-    | None -> false
-    | Some symbolUse ->
+    let symbolUses = reference.ResolveWithFcs(opName, resolveExpr, qualified)
+    symbolUses |> Seq.exists (fun symbolUse ->
```
auduchinok: I wonder if it's safe to change the semantics here. Previously it'd check the first element only, without probing the subsequent ones. I think in some cases we actually wanted to check only the first element, like when there were multiple same-named types with different type parameters count.

auduchinok (commented):
@DedSec256 This is really cool!

I'm not sure about the changes in `resolvesTo`, as some places could actually be relying on checking the first element only. Are you sure it's safe to change it like this? Maybe we should move it to a separate PR?

Please also add more tests. It would be nice to have minimized tests separately covering different type usage kinds.

ReSharper.FSharp/src/FSharp/FSharp.Psi.Intentions/src/Intentions/FunctionAnnotationAction.fs
DedSec256: This is a good reminder. Initially, this case was at the top, but it moved down during logic additions.

ReSharper.FSharp/src/FSharp/FSharp.Psi.Services/src/Util/FSharpResolveUtil.fs
DedSec256: If there are several elements, and the first one didn't fit before this changes, but the list contains the expected symbol, then there is a chance that something didn't work right before?

I can also return `tryHead` in all calls that don't relate to the current PR

ReSharper.FSharp/src/FSharp/FSharp.Psi.Intentions/src/Intentions/FunctionAnnotationAction.fs
DedSec256: I initially looked at AnonRecordTypeDetails but didn't find anything suitable there.

image

ReSharper.FSharp/src/FSharp/FSharp.Psi.Intentions/src/Intentions/FunctionAnnotationAction.fs
DedSec256: Could we just pin type check results before this logic?

auduchinok: Likely yes, but what I'm concerned here is a way to assert that no such checks happen accidentally.

ReSharper.FSharp/src/FSharp/FSharp.Psi.Intentions/src/Intentions/FunctionAnnotationAction.fs
auduchinok: > I initially looked at AnonRecordTypeDetails but didn't find anything suitable there.

Exactly, that is what my comment was about.

ReSharper.FSharp/src/FSharp/FSharp.Psi.Intentions/src/Intentions/FunctionAnnotationAction.fs
```diff
@@ -221,36 +231,61 @@ module SpecifyTypes =
...
+                let fcsSymbol = fcsType.TypeDefinition
+                let declaredElement = fcsSymbol.GetDeclaredElement(context.GetPsiModule()).As<IClrDeclaredElement>()
+                context, reference, fcsSymbol, declaredElement)
+            |> Seq.filter (fun (_, reference, _, declaredElement) -> isNotNull reference && isNotNull declaredElement)
+            |> Seq.toArray
+
+        for context, reference, _, declaredElement in annotationsInfo do
+            reference.SetRequiredQualifiers(declaredElement, context)
+
+        use pinCheckResultsCookie = file.PinTypeCheckResults(true, "Specify types")
```
auduchinok: Would it work if the results are pinned before the modifications?

DedSec256: Do you mean if this code had been called before line 280, or if some logic has already pinned the results before?

auduchinok: Yes, like what is the point of trying to prevent new type checks if we already begin modifications beforehand and force new type checks?

ReSharper.FSharp/src/FSharp/FSharp.Psi.Services/src/Util/FSharpBindUtil.fs
```diff
@@ -13,13 +13,17 @@ open JetBrains.ReSharper.Psi.ExtensionsAPI.Tree
...
+
 // todo: we modify the tree and then ask FCS. What tree is used? Do we wait for the new analysis?
 let bindDeclaredElementToReference (context: ITreeNode) (reference: FSharpSymbolReference)
         (declaredElement: IClrDeclaredElement) opName =
-    reference.SetRequiredQualifiers(declaredElement, context)
 
-    if not (FSharpResolveUtil.resolvesToQualified declaredElement reference true opName) then
-        addOpens reference declaredElement |> ignore
+    reference.SetRequiredQualifiers(declaredElement, context)
+    addOpensIfNeeded reference declaredElement opName
```
auduchinok: Could you explain this change, please?

ReSharper.FSharp/src/FSharp/FSharp.Psi.Intentions/src/Intentions/FunctionAnnotationAction.fs
auduchinok: And maybe we should move this out of this PR and try to do this more systematically instead.

ReSharper.FSharp/src/FSharp/FSharp.Psi.Intentions/src/Intentions/FunctionAnnotationAction.fs
```diff
@@ -14,17 +14,29 @@ open JetBrains.ReSharper.Plugins.FSharp.Psi
...
 open JetBrains.ReSharper.Psi.Tree
 open JetBrains.ReSharper.Resources.Shell
 open JetBrains.ReSharper.Feature.Services.Util
 open JetBrains.Util
 
+[<AutoOpen>]
+module private Extensions =
+    type FSharpMultiplySymbolCandidatesReference(owner) =
+        inherit FSharpSymbolReference(owner)
+        override _.FilterSymbols(symbols) = symbols
```
auduchinok: Can this be in FSharp.Psi project?

ReSharper.FSharp/src/FSharp/FSharp.Psi.Intentions/src/Intentions/FunctionAnnotationAction.fs
DedSec256: The Specify Types code is divided into 5 stages:

1. Adding `ITypeUsage` directly to the tree for parameters/return value.
2. Collecting all named `ITypeUsage`, including recursive ones, i.e. a read-only operation.
3. Getting `IDeclaredElement` and `FSharpSymbolReference` for collected type usages.
4. Adding qualifiers based on previously obtained reference and declared element — modification of the tree (without resolve).
5. Adding required opens: modifies the tree and uses resolves, can repeatedly call a new typecheck for every parameter.

Stage 5 can directly trigger a new resolve, so I pinned the check results before it. 
Stage 4 **doesn't seem to trigger a resolve** (I tried it in tests with ProhibitTypeCheckCookie), so I didn't call it in front of it. 
But I'm not sure that the expected result is guaranteed to be obtained if we pin typecheck results before it, and here's why: after adding a qualifier of at least one type from the parameters, the offset for the current `FSharpSymbolReference` will shift in the tree. It may shift so unsuccessfully that, for example, it matches the name of one of the parameters or initialized properties, and if the type name matches, resolving to this position **may return something completely different from what is expected**, because we have pinned the results of the typecheck before modifying the tree.

ReSharper.FSharp/src/FSharp/FSharp.Psi.Services/src/Util/FSharpBindUtil.fs
DedSec256: I tried to divide the operations into those that trigger resolve and those that don't, so I needed to call the addition of opens separately.  I moved that into a separate function and called it where necessary.

ReSharper.FSharp/src/FSharp/FSharp.Psi.Intentions/src/Intentions/FunctionAnnotationAction.fs
DedSec256: In general, I was not sure that the idea of such an extension through inheritance would be accepted, so at first I left it as an add-hoc implementation. I'll move it to FSharp.Psi, yes.

ReSharper.FSharp/src/FSharp/FSharp.Psi.Intentions/src/Intentions/FunctionAnnotationAction.fs
DedSec256: > And maybe we should move this out of this PR and try to do this more systematically instead.

We can remove typecheck results pinning from this PR, but then I'm not sure what assertions should be added to the code from the previous comments in this PR, because typechecks will occur as needed.

auduchinok: It's fine to have this as is, without the assertions, if we want to add them next. I just want it to make it easier to revert or cherry-pick this PR if needed. Working on such assertions sounds like potentially unrelated changes here and there that would be nice to avoid.

ReSharper.FSharp/src/FSharp/FSharp.Psi.Intentions/src/Intentions/FunctionAnnotationAction.fs
```diff
@@ -35,6 +35,60 @@ module SpecifyTypes =
...
+        match typeUsage with
+        | :? INamedTypeUsage as typeUsage ->
+            let typeReference = typeUsage.ReferenceName
+            let acc = (typeUsage, fcsType, context) :: acc
+
+            let typeArgs = typeReference.TypeArgumentList
+            if isNull typeArgs then acc else
+
+            typeArgs.TypeUsagesEnumerable
+            |> Seq.mapi (fun i typeArg -> i, typeArg)
```
auduchinok:
```suggestion
            |> Seq.indexed
```

ReSharper.FSharp/src/FSharp/FSharp.Psi.Intentions/src/Intentions/FunctionAnnotationAction.fs
```diff
@@ -35,6 +35,60 @@ module SpecifyTypes =
...
+            |> Seq.fold (fun acc (i, typeArg) ->
+                let fcsType = fcsType.GenericArguments[i]
```
auduchinok: What happens if there's a different count of type arguments between the syntax and inferred by FCS?
Let's assume, the type checker can recover incorrect count in the syntax.

ReSharper.FSharp/src/FSharp/FSharp.Psi.Intentions/src/Intentions/FunctionAnnotationAction.fs
```diff
@@ -35,6 +35,60 @@ module SpecifyTypes =
...
+            typeUsage.Items
+            |> Seq.mapi (fun i field -> i, field)
+            |> Seq.fold (fun acc (i, typeUsage) -> collectTypeUsages acc (context, fcsType.GenericArguments[i], typeUsage)) acc
+
+        | :? IParenTypeUsage as typeUsage ->
+            collectTypeUsages acc (context, fcsType, typeUsage.InnerTypeUsage)
+
+        | :? IFunctionTypeUsage as typeUsage ->
+            let fcsType =
+                if fcsType.IsFunctionType then fcsType else getAbbreviatedType fcsType
```
auduchinok: Can we just always call `getAbbreviatedType fcsType`?

ReSharper.FSharp/src/FSharp/FSharp.Psi.Intentions/src/Intentions/FunctionAnnotationAction.fs
```diff
@@ -35,6 +35,60 @@ module SpecifyTypes =
...
+
+            typeArgs.TypeUsagesEnumerable
+            |> Seq.mapi (fun i typeArg -> i, typeArg)
+            |> Seq.fold (fun acc (i, typeArg) ->
+                let fcsType = fcsType.GenericArguments[i]
+                collectTypeUsages acc (context, fcsType, typeArg)) acc
+
+        | :? ITupleTypeUsage as typeUsage ->
+            typeUsage.Items
+            |> Seq.mapi (fun i field -> i, field)
```
auduchinok: Why is this called `field`?

ReSharper.FSharp/src/FSharp/FSharp.Psi.Intentions/src/Intentions/FunctionAnnotationAction.fs
```diff
@@ -35,6 +35,60 @@ module SpecifyTypes =
...
+            let typeUsage = typeUsage.TypeUsage
+            let fcsType = fcsType.GenericArguments[0]
+            collectTypeUsages acc (context, fcsType, typeUsage)
+
+        | :? IAnonRecordTypeUsage as typeUsage ->
+            typeUsage.Fields
+            |> Seq.mapi (fun i field -> i, field)
+            |> Seq.fold (fun acc (i, field) ->
+                let typeUsage = field.TypeUsage
+                collectTypeUsages acc (context, fcsType.GenericArguments[i], typeUsage)) acc
```
auduchinok: It looks like this pattern is repeated at least three times. Would extracting a function and doing a simple `Seq.zip` over type usages and FCS type arguments work here?

ReSharper.FSharp/src/FSharp/FSharp.Psi/src/Resolve/FSharpSymbolReference.cs
```diff
@@ -159,7 +159,24 @@ public FSharpOption<FSharpSymbolUse> ResolveWithFcs([NotNull] string opName, boo
...
+        checkerService.ResolveNameAtLocation(referenceOwner.FSharpIdentifier, names, resolveExpr, opName);
+
+      return FilterSymbols(symbolUses);
     }
+
+    protected virtual FSharpList<FSharpSymbolUse> FilterSymbols(FSharpList<FSharpSymbolUse> symbols) =>
+      symbols.GetSlice(null, 0);
+  }
+
+  public class FSharpMultiplySymbolCandidatesReference(IFSharpReferenceOwner owner) : FSharpSymbolReference(owner)
```
auduchinok: `Multiply`?

ReSharper.FSharp/src/FSharp/FSharp.Psi.Services/src/Util/FcsTypeUtil.fs
```diff
@@ -19,3 +25,7 @@ let getFunctionTypeArgs includeReturnType fcsType =
                 acc
 
     loop fcsType [] |> List.rev
+
+let stripNull (fcsType: FSharpType) =
+    if fcsType.HasNullAnnotation then fcsType.TypeDefinition.AsType()
```
auduchinok: It seems this function does more than stripping the nullability. Does it keep the type arguments properly? If not, let's not add such an API, it's too easy to misuse it.

ReSharper.FSharp/src/FSharp/FSharp.Psi.Services/src/Util/FcsTypeUtil.fs
```diff
@@ -2,6 +2,12 @@
 
 open FSharp.Compiler.Symbols
 
+let rec getFunctionReturnType (fcsType: FSharpType) paramsToSkipCount =
+    if paramsToSkipCount = 0 || not fcsType.IsFunctionType then fcsType else
```
auduchinok, line 6: Does `fcsType.IsFunctionType` work properly with type abbreviations?

ReSharper.FSharp/src/FSharp/FSharp.Psi.Services/src/Util/FcsTypeUtil.fs
```diff
@@ -2,6 +2,12 @@
 
 open FSharp.Compiler.Symbols
 
+let rec getFunctionReturnType (fcsType: FSharpType) paramsToSkipCount =
```
auduchinok: The function name is somewhat ambiguous. Consider `a -> b -> c`. What is the return type? Without the context I'd expect it to be `c`, but here we have another parameter that allows us to get `b -> c` instead. It's not good that one has to guess it from the parameters names instead of using the function name itself.

Can it be something like `skipFunctionParameters`?

ReSharper.FSharp/src/FSharp/FSharp.Psi.Intentions/src/Intentions/FunctionAnnotationAction.fs
```diff
@@ -162,42 +207,72 @@ module SpecifyTypes =
             | :? ITuplePat as tuplePat -> addParens factory tuplePat
             | pattern -> pattern
 
+        let displayContext = displayContext.WithShortTypeNames(true)
```
auduchinok: Any reason to do it repeatedly for all the patterns instead of doing it in the calling code?

ReSharper.FSharp/src/FSharp/FSharp.Psi/src/Resolve/FSharpSymbolReference.cs
```diff
@@ -159,7 +159,24 @@ public FSharpOption<FSharpSymbolUse> ResolveWithFcs([NotNull] string opName, boo
...
 
-      return checkerService.ResolveNameAtLocation(referenceOwner.FSharpIdentifier, names, resolveExpr, opName);
+      var symbolUses =
+        checkerService.ResolveNameAtLocation(referenceOwner.FSharpIdentifier, names, resolveExpr, opName);
+
+      return FilterSymbols(symbolUses);
     }
+
+    protected virtual FSharpList<FSharpSymbolUse> FilterSymbols(FSharpList<FSharpSymbolUse> symbols) =>
+      symbols.GetSlice(null, 0);
```
auduchinok, line 169: What kind of filtering does this function do?

ReSharper.FSharp/src/FSharp/FSharp.Psi.Intentions/src/Intentions/FunctionAnnotationAction.fs
```diff
@@ -162,42 +207,72 @@ module SpecifyTypes =
...
+
+        match annotationsInfo with
+        | [] -> ()
+        | (typeUsage, _, _) :: _ ->
+
+        use pinResultsCookie = typeUsage.FSharpFile.PinTypeCheckResults(true, "Specify types")
+
+        for typeUsage, fcsType, context: ITreeNode in annotationsInfo do
+            let typeReference = typeUsage.ReferenceName
+            let reference = typeReference.Reference.AllowAllSymbolCandidatesCheck()
```
auduchinok, line 265: Do we still need it in this PR?

ReSharper.FSharp/src/FSharp/FSharp.Psi/src/Resolve/FSharpSymbolReference.cs
DedSec256: Takes the first element of the list

auduchinok: Is it correct to call it filtering then? What is the predicate?

ReSharper.FSharp/src/FSharp/FSharp.Psi.Intentions/src/Intentions/FunctionAnnotationAction.fs
DedSec256: This is necessary for cases, for example, when `string` resolves to both the type and the `string` operator (which appears first in a symbol’s list).

ReSharper.FSharp/src/FSharp/FSharp.Psi/src/Resolve/FSharpSymbolReference.cs
DedSec256: The basic idea is as follows:

Our `ResolveNameAtLocation` now uses `GetSymbol Uses AtLocation` for name resolution instead of `GetSymbol Use AtLocation` (which calls `GetSymbolUsesAtLocation` internally and takes the first symbol). Further, at the place of use, where you want to check all symbol uses, you can apply `AllowAllSymbolCandidatesCheck` to a reference, but by default, only the first symbol is checked because of default `FilterSymbols` impl, as if it was called `GetSymbolUseAtLocation` previously.

auduchinok: Thank you, I can see the basic idea in the code. It is, however, unrelated to the question about the filtering.

DedSec256: This logic filters the sequence in such a way that it takes only the element with the 0 index. And `AllowAllSymbolCandidatesCheck` filters it by accepting all elements :)

What do you think would be a better name?

ReSharper.FSharp/src/FSharp/FSharp.Psi.Services/src/Util/FcsTypeUtil.fs
DedSec256: Our `IFunctionTypeUsage a -> b -> c` considers `a` to be the argument type usage and `b -> c` to be the return type usage, for example, hence the naming for this function. But okay, I'll rename it.

ReSharper.FSharp/src/FSharp/FSharp.Psi.Intentions/src/Intentions/FunctionAnnotationAction.fs
DedSec256: Can you show me an example? We get ITypeUsage exactly from FSharpType, in which cases can we have a difference between the syntax and FCS inferring?

ReSharper.FSharp/src/FSharp/FSharp.Psi.Services/src/Util/FcsTypeUtil.fs
DedSec256: This function will not iterate through the parameters of the function abbreviation. If this is an abbreviation, then a top-level logic should decide what to do with it: use it as a single type or get the original function type and process it as a function. 

In some cases, the FCS decides by itself that the parameters from the nested abbreviation should be shifted as parameters of the top-level function, as here

Image

In some cases, the abbreviation itself is enough for us, seems to me

Image

Image

ReSharper.FSharp/src/FSharp/FSharp.Psi.Intentions/src/Intentions/FunctionAnnotationAction.fs
auduchinok: We should make sure that a type reference never resolves to functions.

ReSharper.FSharp/src/FSharp/FSharp.Psi.Services/src/Util/FcsTypeUtil.fs
auduchinok: Ping.

ReSharper.FSharp/src/FSharp/FSharp.Psi/src/Resolve/FSharpSymbolReference.cs
auduchinok: > This logic filters the sequence in such a way that it takes only the element with the 0 index

But it's not filtering then, at least in the terms of how F# defines filtering as checking a predicate on the elements.

ReSharper.FSharp/src/FSharp/FSharp.Psi.Services/src/Util/FcsTypeUtil.fs
DedSec256: This function has been removed from the API

DedSec256: I've added a test for the case you provided. 
`fcsType.TypeDefinition.AsType()` here works as expected with type arguments

https://github.com/JetBrains/resharper-fsharp/blob/ber.a/autoImportSpecifyTypes/ReSharper.FSharp/test/data/features/intentions/specifyTypes/Import%20types%20-%20Type%20with%20null%2003%20-%20Generics.fs


Thank you!

PR #895, opened 2025-07-14
UseDotLambdaSyntaxFix: remove redundant parens (fix)

DedSec256:
The logic is changed back to how it was before a recent formatter update.

image


auduchinok (approved):
Thanks!

PR #957, opened 2025-09-19
Tests: fix, split, update baselines

DedSec256:


auduchinok (approved):
@DedSec256 Thanks a lot!

PR #960, opened 2025-09-21
Separate parserTest into an isolated module

ForNeVeR:
Previously, when running any frontend test in the plugin, IDEA was showing a disambig control, because it didn't know which test is supposed to be run via the `test` task, and which is supposed to run via the `parserTest` task. These were somewhat strangely decided via the test frameworks themselves.

Now, all our tests requiring JUnit are moved to a separate Gradle module (I'll add a corresponding `.iml` file later, after this lands into the monorepo) — this removes any disambig control (because there are no longer two test tasks in the root module), and also allows us to set up the test framework in the `parserTest` module in a more standard way (so we no longer need to wire it up with the main `test` task — both `test` and `check` Gradle lifecycle tasks now will include these tests by default convention, no additional action needed).


auduchinok (approved):

DedSec256 (approved):
Thank you very much!

PR #1126, opened 2026-04-21
Frontend/Psi: avoid extra parsing events by wrapping content in an extra lazy container

DedSec256:
If all the contents of a file are combined into a single `lazy parseable`, then
- only one event is sent during reparsing
- reparsing is reduced to a single `lazy parseable`, and its contents are replaced entirely.

### DIFF
--- a/rider-fsharp/parserTest/build.gradle.kts
+++ b/rider-fsharp/parserTest/build.gradle.kts
@@ -19,6 +19,7 @@ dependencies {
   testImplementation(libs.junit)
   testRuntimeOnly("org.junit.vintage:junit-vintage-engine:5.9.2")
   testRuntimeOnly("org.junit.platform:junit-platform-launcher:1.9.2")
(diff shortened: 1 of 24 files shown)


rider-fsharp/parserTest/build.gradle.kts
```diff
@@ -19,6 +19,7 @@ dependencies {
   testImplementation(libs.junit)
   testRuntimeOnly("org.junit.vintage:junit-vintage-engine:5.9.2")
   testRuntimeOnly("org.junit.platform:junit-platform-launcher:1.9.2")
+  testRuntimeOnly("org.junit.jupiter:junit-jupiter-engine:5.10.2")
```
auduchinok, line 22: Is this change needed?

auduchinok (commented):
Thanks a lot for working on it!

rider-fsharp/parserTest/build.gradle.kts
DedSec256: This is needed to run the parser tests locally, so I added it as a related change

PR #1161, opened 2026-06-16
FSharpSymbolReference/Resolve: support 'unresolved'

DedSec256:


ReSharper.FSharp/src/FSharp/FSharp.Psi/src/Impl/Tree/FSharpFileBase.cs
```diff
@@ -63,6 +63,9 @@ public IReadOnlyList<FcsResolvedSymbolUse> GetAllResolvedSymbols(FSharpCheckFile
     public IReadOnlyList<FcsResolvedSymbolUse> GetAllDeclaredSymbols(FSharpCheckFileResults checkResults = null) =>
       FcsCapturedInfo.GetAllDeclaredSymbols();
 
+    public bool MarkedAsUnresolved(FSharpSymbolReference reference) =>
```
auduchinok: `IsUnresolved` maybe?

ReSharper.FSharp/src/FSharp/FSharp.Psi/src/Resolve/FcsFileCapturedInfo.cs
```diff
@@ -81,6 +81,14 @@ public FSharpSymbol GetSymbol(int offset) =>
     public FSharpDiagnostic GetDiagnostic(Position pos) =>
       GetCachedDiagnostics().TryGetValue(pos);
 
+    public bool MarkedAsUnresolved(FSharpSymbolReference reference)
+    {
+      var resolvedSymbols = GetResolvedSymbols();
+      return reference.SymbolOffset is var symbolOffset && 
+             symbolOffset.IsValid() &&
```
auduchinok: Nit-picking: maybe check the offsets first?

ReSharper.FSharp/src/FSharp/FSharp.Psi/src/Resolve/FcsFileCapturedInfo.cs
```diff
@@ -292,6 +300,18 @@ private ResolvedSymbols CreateFileResolvedSymbols()
         Interruption.Current.CheckAndThrow();
       }
 
+      foreach (var diagnostic in checkResults.Diagnostics)
+      {
+        if (diagnostic.ErrorNumber != 39)
+          continue;
+
+        var identifier = fsFile.GetContainingNodeAt<IFSharpIdentifier>(document.GetTreeStartOffset(diagnostic.Range));
+        var referenceOwner = FSharpReferenceOwnerNavigator.GetByIdentifier(identifier);
```
auduchinok: There may be cases where the diagnostic range also contains the qualifiers. Ideally, we should make it possible to use this logic for other diagnostics too. There's some work being done now to trim the ranges, e.g. in https://github.com/dotnet/fsharp/pull/19818, but we can't rely on the start range yet.

What do you think about finding the tokens the same way it's done for the symbols?

Is there a way to unify it with the cached diagnostics in `FcsFileCapturedInfo` somehow, maybe?

auduchinok (commented):
Cool, thanks!

ReSharper.FSharp/src/FSharp/FSharp.Psi/src/Resolve/FcsFileCapturedInfo.cs
DedSec256: Could you explain, please?

auduchinok: Something like these ideas:

```
if (reference.SymbolOffset is not var symbolOffset || !symbolOffset.IsValid())
    return false;

var resolvedSymbols = GetResolvedSymbols();
return resolvedSymbols.UnresolvedReferences.Contains(symbolOffset.Offset);
```

```
return reference.SymbolOffset is var symbolOffset && symbolOffset.IsValid() && 
        GetResolvedSymbols() is var resolvedSymbols && resolvedSymbols.UnresolvedReferences.Contains(symbolOffset.Offset);
```

```
return reference.SymbolOffset is var symbolOffset && symbolOffset.IsValid() && 
        GetResolvedSymbols().UnresolvedReferences.Contains(symbolOffset.Offset);
```

ReSharper.FSharp/src/FSharp/FSharp.Psi/src/Impl/Tree/FSharpFileBase.cs
DedSec256: I initially wanted to name it that way, but I suppose it might be confusing. For example, all references without a symbol might be assumed to fall into this category.

auduchinok: > For example, all references without a symbol might be assumed to fall into this category.

Why? We check that no symbol situation also corresponds to the error, don't we?

DedSec256: image

I'd expect
```fsharp
IsUnresolved(A) = true

IsUnresolved(B) = true
```
since both references are without the symbol


```fsharp
MarkedAsUnresolved(A) = true

MarkedAsUnresolved(B) = false
```
since only the first one is marked as unresolved by FCS


As for me, `IsUnresolved` implies broader semantics. If you don't think so, I can rename it.

ReSharper.FSharp/src/FSharp/FSharp.Psi/src/Resolve/FcsFileCapturedInfo.cs
DedSec256: >There may be cases where the diagnostic range also contains the qualifiers.

Is it really possible for Unresolved-diagnostics to return a range with a qualifier? Currently, we do not perform any additional processing such as https://github.com/JetBrains/resharper-fsharp/blob/3be976aadb3f50d908bdd7a61973efbc9d354eeb/ReSharper.FSharp/src/FSharp/FSharp.Psi/src/Resolve/FcsFileCapturedInfo.cs#L314
 in `FcsErrorsStageProcessBase` 

> Is there a way to unify it with the cached diagnostics in FcsFileCapturedInfo somehow, maybe?

As I can see, only errors that don't care about idents are cached, like TypeEquation or FunctionValueUnexpected. I'm not sure it could be unified properly.

ReSharper.FSharp/src/FSharp/FSharp.Psi/src/Impl/Tree/FSharpFileBase.cs
auduchinok: I see, thanks!

To me "marked" implies that there's a place where we mark it explicitly somewhere, and I wanted to deemphasize this word. Maybe it could be something like `IsIgnored` instead (i.e. when not marked) or `ShouldReportError` (to explain the reason to have this method), or something along these lines?

ReSharper.FSharp/src/FSharp/FSharp.Psi/src/Resolve/FcsFileCapturedInfo.cs
auduchinok: > Is it really possible for Unresolved-diagnostics to return a range with a qualifier?

Something related to indexers, ranges/spans used to or still returns some weird ranges.

ReSharper.FSharp/src/FSharp/FSharp.Psi/src/Impl/Tree/FSharpFileBase.cs
DedSec256: maybe `IgnoreIfUnresolved` or `ShouldReportResolveError`?

auduchinok (approved):

PR #1165, opened 2026-06-18
Import type: add a popup action

DedSec256:
Currently, this only works for importing types, as importing extensions and module members requires additional infrastructure work to allow multiple imports.


auduchinok (approved):
This is really cool, thanks!

PR #1212, opened 2026-08-03
FcsCapturedInfoCache: support forks

DedSec256:
- [x] Add a test


DedSec256:
image

ReSharper.FSharp/src/FSharp/FSharp.Psi/src/Resolve/FcsCapturedInfoCache.cs
```diff
@@ -24,18 +25,26 @@
...
     private readonly FSharpScriptPsiModulesProvider myScriptPsiModulesProvider;
     private readonly IShellLocks myLocks;
-    public IFcsProjectProvider FcsProjectProvider { get; }
 
-    protected readonly Dictionary<FcsProjectKey, FcsModuleCapturedInfo> ModuleCaches = new();
-    protected readonly Dictionary<IPsiModule, FcsModuleCapturedInfo> ScriptCaches = new();
+    private readonly LazyForkedCopyOnWriteContentModelData<State> myState = new
+    (
+      factory: static () => new State(),
+      copyFactory: static _ => new State()
```
controlflow, line 42: I'd suggest to comment here on why exactly we are not making a copy here. We only need the reuse off the non-forked data during the Read operations, and on write operations we are find here with this cache becoming empty. It will be computed on-demand.

ReSharper.FSharp/src/FSharp/FSharp.Psi/src/Resolve/FcsCapturedInfoCache.cs
```diff
@@ -195,35 +213,40 @@ public void OnPsiChange(ITreeNode elementContainingChanges, PsiChangedElementTyp
 
     public void SyncUpdate(bool underTransaction)
     {
-      using var writeCookie = WriteLockCookie.Create(underTransaction);
-      using var lockCookie = MonitorInterruptibleCookie.EnterOrThrow(mySyncObj);
+      if (myState.ValueForRead.DirtyFiles.IsEmpty())
+        return;
```
controlflow, line 224: I think this is wrong. It will always be empty, since 1) forks are usually created when the PSI is fully committed and the caches are updated, 2) there can be no modification happening outside of the fork, since fork lives under the Read lock

ReSharper.FSharp/src/FSharp/FSharp.Psi/src/Resolve/FcsCapturedInfoCache.cs
```diff
@@ -235,56 +265,111 @@ public bool HasDirtyFiles
...
+    private IFcsFileCapturedInfo GetOrCreateProjectFileCapturedInfo(IPsiSourceFile sourceFile)
+    {
+      var psiModule = sourceFile.PsiModule;
+
+      if (psiModule.ContainingProjectModule is not IProject)
+        return EmptyFcsFileCapturedInfo.Instance;
+
+      var projectKey = FcsProjectKey.Create(psiModule);
+
+      {
```
auduchinok, line 328: Is this block needed?

auduchinok (approved):
This is incredible work. I've started doing almost the same but then scrapped it off half way due to how FCS state could get corrupted and I didn't know how to proceed. Let's hope it's not too bad 🙂

ReSharper.FSharp/src/FSharp/FSharp.Psi/src/Resolve/FcsCapturedInfoCache.cs
DedSec256: Here, two shorthands are used: the variables stateForRead and stateForWrite. This block is the most non‑invasive way to limit access to stateForRead so that you don’t accidentally use it instead of stateForWrite below (I did it once, lol).

Do you know how we can do this better?

auduchinok: I don't mind keeping it as is, I was just wondering if there was a technical reason that I couldn't spot.

PR #1221, opened 2026-08-18
FcsCapturedInfoCache: allow to calculate a shared data within read-only forks

DedSec256:


ReSharper.FSharp/src/FSharp/FSharp.Psi/src/Resolve/FcsCapturedInfoCache.cs
```diff
@@ -336,7 +342,7 @@ private IFcsFileCapturedInfo GetOrCreateProjectFileCapturedInfo(IPsiSourceFile s
         }
       }
 
-      var stateForWrite = myState.ValueForWrite;
+      var stateForWrite = IsReadonlyFork ? myState.ValueForRead : myState.ValueForWrite;
       lock (stateForWrite)
```
DedSec256, line 339: Perhaps it would be better to use MonitorInterruptibleCookie.EnterOrThrow(state) instead of lock here?

DedSec256 (commented):
There may be a situation, for example, during SWA for C#, when a readonly fork attempts to access a resolve data that has not yet been computed. The readonly fork may perform computation and write it in the shared state — logically, this is safe, since the readonly fork will not change this data in the future; the data will remain in the cache and will be available outside the fork. How much should we worry about the growth of this cache? It seems that it will be not worse than it was before.

ReSharper.FSharp/src/FSharp/FSharp.Psi/src/Resolve/FcsCapturedInfoCache.cs
```diff
@@ -52,6 +52,12 @@ protected sealed class State
 
     public IFcsProjectProvider FcsProjectProvider { get; }
 
+    /// Read-only forks are used by SWA and can compute missing FCS data accessible outside the fork
+    /// since they cannot modify it further.
+    private static bool IsReadonlyFork => 
+      ContentModelFork.IsCurrentlyForked &&
+      !ContentModelFork.CurrentCapabilities.HasFlag(ContentModelForkCapabilities.WriteOperations);
```
controlflow, line 59: HasFlag is allocation heavy (for what it do) on .NET Framework runtime, `&` + `!= 0` is better

DedSec256: HasFlag is used in the platform. I think we can fix it separately later throughout all its usages.

